PR #483, opened 2015-07-03
Add dash variations to symbols.xml

simon-v:


simon-v:
Since there is one such string for every language, this may be a slightly less trivial change. I'll leave it for when i can do it properly.

simon-v:
PR #487 makes this one redundant. Please consider merging that instead.

PR #487, opened 2015-07-10
Add dash variations and square root symbol to symbol tables

simon-v:
Adding the square root symbol required moving the division symbol to the multiplication symbol key to make room.

### DIFF
--- a/src/main/res/xml/symbols.xml
+++ b/src/main/res/xml/symbols.xml
@@ -39,7 +39,7 @@
         <Key android:codes="58" android:keyLabel=":" android:popupCharacters=";"/>
         <Key android:codes="63" android:keyLabel="\?"/>
         <Key android:codes="47" android:keyLabel="/"/>
-        <Key android:codes="45" android:keyLabel="-"/>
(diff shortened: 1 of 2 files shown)


menny:
Hi, @simon-v , what are you plans with this?

simon-v:
Eventually? To extend the symbols keyboards to include every commonly and semi-commonly used symbol, perhaps using PerfectKeyboard as an example.

Right now, though, i *think* i got all of the bugs with this contribution. The build on Shippable seems to fail during the install phase, not the build phase. There isn't much i can do about that.

simon-v:
1. I can't figure out what Shippable is complaining about. All five XML validators i tried seem to think the file is well-formed, and Shippable itself doesn't disclose the exact nature of the complaint.
2. I can't figure out how to build the project locally, despite following the setup and build instructions very carefully.
3. I can't help but feel extremely frustrated about both of those.

menny:
I'm really sorry about that - it seems that Shippable is broken.

Can you squash all those commits into one, and force push to this PR. I'll merge even if shippable is failing.

simon-v:
I think i got it.
I'd prefer if a few simple quality tests were made after merging in case i'm missing something.

Issue #506, opened 2015-09-16
Gesture "Swipe up" does not work for "Toggle uppercase"


pan2003:
Using latest version 1.5-r4.1.

In settings, I assigned gesture "Swipe up" to "Toggle uppercase". Does not work. Also tried to assign the same action to "Swipe up from space", does not work either.

But "Swipe up from space" for action "Utility keyboard" works fine. And other gestures (toggle languages at swipe right, toggle symbols keyboard at swipe left, hide keyboard at swipe down) works fine.

menny:
Because we are handling the SHIFT state only using the `onPressed` and `onRelease` calls, we are not doing anything when `handleShift` is called.

pan2003:
It works strange in latest AnySoftKeyboard-debug-1.6.92-all-beta2.apk.

The keyboard is now in lowercase state. I swipe up. Now I'm in UPPERCASE state. That's ok.
But now I swipe up to shift back to lowercase state. But I still see UPPERCASE letters.

After some testing I found that when I start to swipe up the keyboard really toggles back to lowercase, but when I continue swiping in shifts to UPPERCASE again.

pan2003:
Just tried AnySoftKeyboard-release-1.6.97-all.apk. Behaves the same.
I'll open a new bug.

Issue #502, opened 2015-09-09
Long press SHIFT should move to caps-lock


menny:

pan2003:
As far as I understand, now CAPS LOCK can be toggled by SHIFT double click. So probably this enhancement is not of high priority...

pan2003:
By the way, after fixing bug #506 about gesture for SHIFT... If I swipe up to toggle SHIFT and than swipe up again, what will happen? Here could be 2 variations:
1) SHIFT is toggled between 2 states: on and off,
2) SHIFT is toggled between 3 states: on, CAPS LOCK, off.

menny:
Option 3: it will cycle between ON->LOCK->OFF (just as if you clicked SHIFT)

menny:
Going to revert this one... this behavior breaks the _hold shift while pressing other keys to get caps, then release shift to return to un-caps state_..

PR #515, opened 2015-10-21
Add lean light theme.

danielegobbetti:
Most of the graphics were with imagemagick using convert -negate on the dark graphics.

There is a problem with the smiley Key, since it's defined in an XML file that is not
theme dependent, as far as I can tell. The smiley key is hence currently invisible
but working.


menny:
Thanks!
It looks good on my device. I'm looking into the smiley icon problem.

Can you help one with another task?

danielegobbetti:
I can try, which task are you referring to?

Thanks for the fast merge  :)

menny:
#506 or #502 are good options

danielegobbetti:
Well, I'll try but I don't know if I'll be able to help, the codebase is pretty big and I just worked on the res files so far.

I'll try to invert the gradient on the keyboard to workaround the smiley issue, shall I do a new pull request of can I just attach a patch here? It should literally be a two-line change.

menny:
Regarding the theme, open a new PR. Also, can you look at removing the
underline brackets from the function keys (shift, space, etc)

PR #517, opened 2015-10-23
Lean light theme refinements

danielegobbetti:
Inverted the background gradient to make the smiley at least visible.
Removed the underline brackets from function keys as per upstream request.


menny:
It looks great.
A request: can you update the screenshot too?
One more suggestion: The popup background color is cool, but I think that the _normal_ popup-key's background should not be white, it should be transparent. I think.

danielegobbetti:
Hi I will probably be offline till Monday, then I'll happily do the modifications you suggest. 
The popup keys are already white, and the selected popup key is light orange, so I'm not sure I understand that point fully, could you please elaborate more?
--

PR #521, opened 2015-11-03
Add em dash to the regular bottom row

andrewshadura:


menny:
Can you also add _em dash_ to `ext_kbd_bottom_row_ng.xml`?

menny:
I see that it already has that character. Forget about it then.

Thanks for helping out.

PR #685, opened 2016-07-29
fixed two fingers swipe, and also added option for swipe up and down …

mtudora:
…#252


menny:
thanks for this.
I'll pull this and try it out

Issue #716, opened 2016-09-28
Google Voice Input missing dialog should not be interactive


marmistrz:
### Steps to reproduce
1. Tap the microphone icon without Google Voice Input installed 

### Expected behaviour
A non-modal toast notification should be displayed. The dialog should not require interaction or steal focus.

### Actual behaviour
A modal dialog requiring user's attention is displayed. It's irritating for those who don't want Google's proprietary software on their phones.

**Android OS version:**
CM 13.0 without Google Apps

djibux:
What about having the possibility to disable and hide the voice input button completely? I could use a slightly wider spacebar :)

Kaljurand:
Related cases:

- https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/1672
- https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/927

PR #750, opened 2016-10-24
Use a non-blocking toast to notify about missing voice input. Closes …

marmistrz:
…#716


marmistrz:
Can you please explain why the check has failed? I can see nothing interesting in Details.

menny:
it fails because you changed code but did not cover it with a unit-test

app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -1350,9 +1349,12 @@ private void onFunctionKey(final int primaryCode, final Key key, final int multi
                 if (mVoiceRecognitionTrigger.isInstalled()) {
                     mVoiceRecognitionTrigger.startVoiceRecognition(getCurrentAlphabetKeyboard().getDefaultDictionaryLocale());
                 } else {
-                    Intent voiceInputNotInstalledIntent = new Intent(getApplicationContext(), VoiceInputNotInstalledActivity.class);
-                    voiceInputNotInstalledIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-                    startActivity(voiceInputNotInstalledIntent);
+                    Context context = getApplicationContext();
```
menny: I think that this will be very confusing for non-power-users:
They will get a _toast_ and will not know what to do to get voice functionality.

This is the main reason that Activity exists - it give information and have a button that direct the user to the Play-Store to download the voice-input app.

menny:
I can suggestiopn a different approach:
show the VoiceInputNotInstalledActivity, but add another button that says "Don't notify me again".

app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
marmistrz: Not all users have Play Store installed. This information could be contained in README, Help, whatever. But I agree that the Toast should mention that the information is in README

marmistrz:
This is an option

PR #807, opened 2016-12-24
Implement experimental gesture typing

justinmichaud:
Solves issue 264

TODO:
- Undo making some of the members public
- I have no idea how to make the corrections work properly, or how to get the data I need for automatic capitalization and spacing (see AnyKeyboardView onTouch)
- I added a wordlist txt file, because I couldn't figure out how to get a list of words from the native dictionary
- I have no clue how to access the character associated with a key (I was getting weird null errors), so I used label instead. This is obviously a horrible hack
- It currently checks that you started your gesture near a key with a latin a-z or A-Z key to see if gesturing is enabled. Also a massive hack (see GestureTypingDetector isValidStartTouch)
- Enhance gesture detector: My plan was to implement something similar to the linked paper in the issue, but it looks like that will require a linear scan through all of the words in the dictionary. For each word, we can compute a probability that it is the correct word, and then sort the words based on this probability.


jtjj222:
To avoid a linear search, we can group the words based on proximity to the first and last letter. See [page 21 of the paper](https://esc.fnwi.uva.nl/thesis/centraal/files/f2109327052.pdf). I think we should sort out the other issues first. I'm going to work on implementing the path simulation described in the paper later today.

menny:
@jtjj222 I did some moving. I believe I got rid of most of the public members and horrible class-couplings.

This is still very rough, of course, I still have a few ideas to implement: I'll probably want to move more logic into the new `AnySoftKeyboardWithGestureTyping` class, and make the _gesture-typing-started` state detection much cleaner.
I also want to add a bunch of unit-tests, but let's do that after we realizing what is the API we want to agree.

Last thing: I did some research on how to get the list of words from the JNI dictionary, and it's not easy, nor is it memory efficient. I'm sorry to say, but we'll have to extend the JNI dictionary and implement the method `GestureTypingDetector.getGestureWords` there.

ywum7fhv0skqh07h:
Is this still too early development to be reporting bugs?

I'm happy to list what I have observed so far, unless there's too many temporary hacks which will drastically change behaviour when rewritten.

justinmichaud:
See #1040

PR #836, opened 2017-01-06
Android automatic refactor - Recycle

luiscruz:
Hi,

I am developing a tool to automatically refactor Android applications with the goal of improving energy efficiency.
This pull request has the changes generated while applying the rule "Recycle".

Some resources (e.g., ```Cursor``` instances) should be closed when they are no longer necessary. 

I have made a previous validation of the changes and they seem correct.
Please consider them and let me know if you agree with them.

Best,
Luis


menny:
@luiscruz will you have additional suggestions in the future? I'm happy about those.

luiscruz:
Thanks for the feedback @menny !

I'm currently writing a tool demo that will explain this in detail. It will be online soon.
I'll try to add more checks in the near future.
If you know optimisations for energy efficiency that you could recommend I'd be happy to hear from you!

I've just removed the ```null``` checks.
Right now checks are pending, but previously there were 2 checks that did not pass.
What was the issue?

menny:
@luiscruz other than running linters and unit-tests checks (which your change passed), there is also two other checks that verify that your change includes unit-tests and that it does not reduce the overall test coverage. You shouldn't worry about that, the code you touched is in an area that does not need unit-tests, so it's fine.

PR #877, opened 2017-01-31
Reorder right parenthesis and brackets

steelman:
Make the order of right parenthesis, square and curly brackets match the
order of left ones, the same way '<' and '«' match their counterparts.


menny:
Nice catch. Thanks.

menny:
You're change will be available in the next release. Also, in the next beta which will be available later today

steelman:
I think there are more patches like this to be made. For example when long-pressing the upper row keys (qwerty...), the digits (0-9) aren't always the highlighted by default. Are you interested in such fixes?

menny:
@steelman of course, anything you think can make the keyboard better.

PR #887, opened 2017-02-06
Add xml-eu folder

aldatsa:
The xml-eu folder includes the following new files:
- popup_domains.xml
    Changes:
    - Customised list of top level domains.
- currency_data.xml
    Changes:
    - Show the euro sign (€) by default, instead of the dollar sign ($).
- ext_kbd_bottom_row_ng.xml
    Changes:
    - Show hyphen (-) by default, instead of single quote (').
- ext_kbd_bottom_row_regular.xml
    Changes:
    - Show hyphen (-) by default, instead of single quote (').
- ext_kbd_bottom_row_regular_with_voice.xml
    Changes:
    - Show hyphen (-) by default, instead of single quote (').
    - Show the colon symbol (:) before the semicolon symbol (;) on the popup.


app/src/main/res/xml-eu/ext_kbd_bottom_row_ng.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="utf-8"?>
```
menny: I'm not sure this is the best way to do this. I sometimes change those bottom rows, and I plan to add another bottom row. I most likely forget to change the Basque-specific rows.

menny: What exactly are you trying to achieve with those bottom-row layouts?

aldatsa: If you think that the changes to the bottom rows are going to be problematic in the future I can remove them from the pull request. I don't want to cause you any inconvenience.
The hyphen and colon symbols are more used in Basque than the single quote and the semicolon, specially the hyphen, as it's used to add suffixes and to create compound words. We thought that it's more useful for Basque speakers that way.

menny: In that case, let me suggest this:
Instead of altering the existing bottom rows, you may want to create an additional bottom row  with hyphen.

I would also explore the possibility to add all those bottom rows as a part of the Basque language pack.

menny:
I'm going to wait for this PR to be merged before releasing v1.8-r9.

app/src/main/res/xml-eu/ext_kbd_bottom_row_ng.xml
aldatsa: Ok, thanks for your help. In your opinion, which one is the best option? Should I create an additional bottom row with hyphen or should I wait until you explore the possibility to add all those bottom rows as a part of the Basque language pack?

menny: I think it would make much more sense to have those coming from the language pack.

menny:
@aldatsa what do you want to do with this PR?

aldatsa:
@menny One option is to merge the non-problematic files (currency_data.xml and popup_domains.xml) and close this pull request. Is it possible to cherry-pick these files?
That way the users have something usable while we think what to do with the bottom rows.

menny:
@aldatsa The best we to do that is for you push another commit into this branch that removes the problematic files.

aldatsa:
Done!

menny:
thanks

Issue #966, opened 2017-04-05
Dvorak and Colemak variants


asterismo:
Can you add those?

menny:
We are working on this.
See https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/902

Issue #902, opened 2017-02-19
Migrate alternative layouts to main APK


menny:
colemak
dvorak (also include ń)
qwerty_compact
terminal

menny:
@friesenkiwi interested in doing this?

friesenkiwi:
@menny Yes, actually this is a dupe of https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/748, also https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/422 and https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/579 are related.

I did want to take a stab at those, as soon as the LanguagePacks are all finished ;-)
Need to try to compile AnySoftKeyboard first, never did anything with the NDK.

menny:
@friesenkiwi let me know how I can help, but I'm going to guess that if you have NDK installed, you wouldn't have any problems compiling ASK

friesenkiwi:
Ok, good to know! :-)

asterismo:
👍

friesenkiwi:
OK, after installing the NDK, the build went smooth, so I should actually be able to work on this issue.

menny:
Nice!

asterismo:
Dvorak spanish variant also swaps the R with the H, can't wait to have it! Please, upload it to FDroid asap!

friesenkiwi:
@asterismo What do you mean "swaps the R with the H"? Is there something to be changed in the Spanish pack or  to be considered when migrating AlternativeEnglishLayouts?

menny:
Thank you @friesenkiwi.

menny:
A beta release will be available later today for testing

menny:
Actually, it will be available as alpha in a few minutes

asterismo:
Spanish Dvorak variant changes places of H and R besides adding the Ñ. It is default now in most linux distros, https://es.m.wikipedia.org/wiki/Teclado_Dvorak It really should be considered because there will be a difference between the PC and phone... I have installed an extra APK from GPlay named Alternative Layouts which is not available in FDroid... this should be bundled with the base app.

friesenkiwi:
@asterismo the Dvorak and other keyboards from the AlternativeLayouts app are now merged into the main app (that's the purpose of this issue), and will be available in the next release.
[...]

Issue #748, opened 2016-10-24
Please import the alternate layouts to GitHub


marmistrz:
The Dvorak and Colemak keyboards used to be in the SVN repo: http://softkeyboard.googlecode.com/svn/trunk/LanguagePacks/AlternativeEnglishKeyboards/
which is no more

This may refer to all your language packs. So please import everything that was there :)

Thanks

menny:
Will track progress with #902

Issue #422, opened 2015-03-24
Polish Dvorak layout missing a letter


antekone:
I can't seem to find the proper place for this, so I'm sorry if I'm doing it wrong.

I'm using ASK + Dvorak layout with Polish language, but the "ń" letter is missing (together with "Ń"). It's a standard letter in Polish alphabet, and I think it should be included. Do you think it's possible to include it? Can I do it myself? (changing the order of the alternative letters also would be nice)

menny:
Which Polish language pack are you using? Can you send me the link?

antekone:
Sorry for my delay.

Actually I was wrong, I'm not using Polish layout at all. I'm using [Alternative English Language Pack](https://play.google.com/store/apps/details?id=com.anysoftkeyboard.languagepack.alt.english), which includes Dvorak layout. I was confused, because this English pack has most of Polish letters, with a sole exception of "ń" and its uppercase variant. Sorry for that.

Anyway, is it possible to include "ń" in this pack?

menny:
Will track progress with #902

PR #967, opened 2017-04-10
Importing AlternativeEnglishLayouts (Colemak, Dvorak, compact) and SSH/Terminal

### LINKED ISSUE #579: (Re)assignment of Polish diacritics in English 16 key layout
Hello,
would it be possible to assign letter ą to _"abc"_ key's popup in 16 key English layout and reassign ł to key _"jkl"_ instead of _"ghi"_? The former would make it possible to use all Polish diacritics with that layout and I think the latter would make more logical sense considering ł is a derivative of l (and not i).

friesenkiwi:
Closes #579,  #422, #748, #902, #966

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -367,11 +367,11 @@
     <string name="symbols_numbers_keyboard">123</string>
     <string name="symbols_time_keyboard">time</string>
     <string name="eng_keyboard">English</string>
-    <string name="dvorak_keyboard">Dvorak</string>
-    <string name="compact_keyboard">Compact QWERTY</string>
-    <string name="compact_keyboard_16keys">16 Keys English</string>
+    <string name="dvorak_keyboard">EN Dvorak</string>
+    <string name="compact_keyboard">EN Compact</string>
+    <string name="compact_keyboard_16keys">EN Keypad</string>
     <string name="terminal_keyboard">Terminal</string>
-    <string name="colemak_keyboard">Colemak</string>
+    <string name="colemak_keyboard">EN Colemak</string>
     <string name="keyboard_change_locked">n/a</string>
 
     <string name="selected_add_on_summary">Now using \'<xliff:g id="addOnName">%s</xliff:g>\'
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -367,6 +367,11 @@
     <string name="symbols_numbers_keyboard">123</string>
     <string name="symbols_time_keyboard">time</string>
     <string name="eng_keyboard">English</string>
+    <string name="dvorak_keyboard">Dvorak</string>
+    <string name="compact_keyboard">Compact QWERTY</string>
+    <string name="compact_keyboard_16keys">16 Keys English</string>
+    <string name="terminal_keyboard">Terminal</string>
+    <string name="colemak_keyboard">Colemak</string>
     <string name="keyboard_change_locked">n/a</string>
 
     <string name="selected_add_on_summary">Now using \'<xliff:g id="addOnName">%s</xliff:g>\'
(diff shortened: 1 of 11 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -367,6 +367,11 @@
     <string name="symbols_numbers_keyboard">123</string>
     <string name="symbols_time_keyboard">time</string>
     <string name="eng_keyboard">English</string>
+    <string name="dvorak_keyboard">Dvorak</string>
+    <string name="compact_keyboard">Compact QWERTY</string>
+    <string name="compact_keyboard_16keys">16 Keys English</string>
```
menny: This should be much shorter since this text will also be shown on the keyboard's language-switch key.
Maybe "T9 English"? Or "16 Keys EN"?

menny:
great work @friesenkiwi

friesenkiwi:
Updated the strings and the two of the three test occurences, the first one seems to be a bit more tricky, see https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/0293796132ed78ab15bbff0857b9b66416a462a9/app/src/test/java/com/anysoftkeyboard/AnySoftKeyboardKeyboardSubtypeTest.java#L60

``` 
        InputMethodSubtype[] reportedSubtypes = subtypesCaptor.getValue();
        Assert.assertNotNull(reportedSubtypes);
        Assert.assertTrue(reportedSubtypes.length > 0);
        Assert.assertEquals(keyboardBuilders.size(), reportedSubtypes.length);
        for (int builderIndex = 0; builderIndex < keyboardBuilders.size(); builderIndex++) {
            KeyboardAddOnAndBuilder builder = keyboardBuilders.get(builderIndex);
            InputMethodSubtype subtype = reportedSubtypes[builderIndex];
            Assert.assertEquals(builder.getKeyboardLocale(), subtype.getLocale());
            Assert.assertEquals(builder.getId(), subtype.getExtraValue());
        }
```
I'm not sure how to handle this here, since unfortunately I am not too familiar with the test framework, sorry :-( Can you suggest something?

menny:
@friesenkiwi I would change this code to:
```
InputMethodSubtype[] reportedSubtypes = subtypesCaptor.getValue();
        Assert.assertNotNull(reportedSubtypes);
        Assert.assertEquals(7, keyboardBuilders.size());
        Assert.assertEquals(6, reportedSubtypes.length);
        int reportedIndex = 0;
        Assert.assertEquals(keyboardBuilders.size(), reportedSubtypes.length);
        for (KeyboardAddOnAndBuilder builder : keyboardBuilders) {
            if (TextUtils.isEmpty(builder.getKeyboardLocale())) continue; //Terminal does not have a locale
            InputMethodSubtype subtype = reportedSubtypes[reportedIndex++];
            Assert.assertEquals(builder.getKeyboardLocale(), subtype.getLocale());
            Assert.assertEquals(builder.getId(), subtype.getExtraValue());
        }
        Assert.assertEquals(reportedIndex, reportedSubtypes.length);
```

If this does not work, you can mark the test as `@Ignore` and I'll fix that up.

friesenkiwi:
Ah, OK, that's quite pragmatic :-)
Did the changes and now all tests succeed at least on CircleCI. So all seems good?

I tried to run the tests locally (by running the Gradle Task AnySoftKeyboard->AnySoftKeyboard (root)->Tasks->verification->testReleaseUnitTest in AndroidStudio), but heaps of tests fail because of
`    java.lang.UnsatisfiedLinkError: Can't load library: /tmp/robolectric5423916269313465430/app_lib/libanysoftkey2_jni.so.1.0`
I checked /tmp and yes, there are a couple of folders and files created during run, but they are all empty. Do I need to run another task beforehand to prepare?
I also tried to run the specific unit test separately, but that also fails because of missing system properties (I think they are set in the Gradle file). Is there a way to set those properties before running a single test?

PR #1000, opened 2017-05-28
Create variant of Yochees Dark theme with hints

nvllsvm:
Adds a variant of Yochees Dark theme with hints. Similar to Lean Dark - Option 2

screenshot_20170528-155000


menny:
Thanks for that.

menny:
I'm researching why it fails. It shouldn't

PR #1003, opened 2017-06-01
Added code to prevent users or (buggy apps) take screenshots

ranjeetDT:
Added FLAG_SECURE to all the Activities so that no screenshots can be captured. For further information about FLAG_SECURE, read [FLAG_SECURE](https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#FLAG_SECURE) .

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
+++ b/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
@@ -17,6 +17,7 @@
 package com.anysoftkeyboard;
 
 import android.annotation.SuppressLint;
+import android.app.Activity;
 import android.content.Context;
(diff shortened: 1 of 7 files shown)


menny:
why would we want that?

menny:
I'm closing this since I do not see a reason.

ranjeetDT:
@menny 
FLAG_SECURE is a flag is used to prevent any user or screen-capturing app to capture device's screen. This is useful in scenarios where there is any sensitive activity such as password keystrokes, credit card data, etc, it will prevent capture of anything thus preventing compromising the user's data.
In a keyboard app like yours, this seems to be a useful thing to do.
Below is a demo of the same:
https://www.youtube.com/watch?v=tT1XSoykjtA

PR #1004, opened 2017-06-04
fix german translation

mueller-ma:


menny:
Thanks for this.
Translations are done via https://crowdin.net/project/anysoftkeyboard and not through PRs.

Can you make the same fix at https://crowdin.net/project/anysoftkeyboard ? (I can do that too, if you prefer).

mueller-ma:
Can you do it? I would have to create an account first.

Can you also add a note to README that translations are done via this website?

PR #1026, opened 2017-06-21
Add arrow symbols as popups to ^ on symbols board

nwf:


menny:
Please attach a screenshot of this change

nwf:
Oi.  That's going to take some time, as I will have to build and install the mod.  Am I missing something which makes this change insufficient?

menny:
The thing that I'm concerned about it that users will think these are actual arrows action keys, and not just printable arrows. I want to see how this looks.

nwf:
OK, gimme a while to get studio happy (apparently I'm several updates behind).

For whatever it's worth, these symbols are on ^ on gboard, I think, and I was surprised to find the actual cursor control keys in ASK but not the arrow symbols.  (Perhaps it'd be better to add them next to the other arrows on the emoji pages, if you think on ^ is likely to be confusing.  I'd suggest adding them to the arrow keys themselves, but they're tagged as repeating and so push-and-hold seems unlikely to work out. :))

nwf:
screenshot

Screenshot attached.

menny:
Thank you

menny:
This change will be in ALPHA channel in an hour or so

PR #1029, opened 2017-06-24
Emoji o

menny:


brevilo:
I'm curious: should this enable the latest emojis on Nougat (7.1.2) as well, at least when `NotoColorEmoji.ttf` contains the relevant characters?

Thanks

PR #1040, opened 2017-07-02
Implement experimental gesture typing

justinmichaud:
Solves #264.

TODO:
[ ] Auto capitalization + selecting a suggestion work properly

After this is merged, next steps would be to examine factoring in a language model to the word distance metric, and to consider implementing something like [LONG SHORT TERM MEMORY NEURAL NETWORK FOR KEYBOARD GESTURE DECODING]( https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/43461.pdf) instead of the naive algorithm used now.


menny:
You did it!!!!

Vynikal:
Awesome work. Is there a way to make an external lib to be used with AOSP Keyboard? Just like libjni_latinimegoogle.so? I miss an open source alternative.

justinmichaud:
It would be possible, although I don't think it would share much code with this. I will take a look in January when I am on co-op

alterechtschreibung:
@justinmichaud Thanks. 
Reference: https://github.com/microg/android_packages_apps_GmsCore/issues/460

IzzySoft:
@justinmichaud mind mentioning which January? :wink: Kidding. Just a reminder.

menny:
@IzzySoft this is only me blocking gesture-typing rn.
But, worry not! I've started implementing optimization and verifications so it can be added to the production app.

IzzySoft:
Cool! Looking forward to have that right in LOS – and FOSS :tada:

PR #1065, opened 2017-07-19
Add a few additional short smileys

jgkamat:
This PR adds support for `:D` `B)` and `:O`, which are fairly common smileys (to me, at least) but missing form the stock text smileys. 

I also had to make a small change to get this to compile at all, let me know if that shouldn't be added and I'll trash that commit.

One more point, shouldn't the 'long form' of smileys print out long smileys? Pressing on `:-)` prints out `:)` on the debug build (and that makes sense, since that's how it's listed out in the field), but I have a phone with the stable release that /sometimes/ works? I'm not really sure what's going on there. It doesn't matter too much to me, because shorter smileys rule `:)`

Let me know if anything should be changed, or if you spot anything amiss.


menny:
thanks.

I think that long-form is a bug.

PR #1185, opened 2017-12-05
Adding en subtype to method.xml

menny:


menny:
@phhusson this is the change I mentioned.

PR #1186, opened 2017-12-07
HintText contains substring if it's too long.

avently:
Now if hintText is too long it is replaced with dots. I think we need to show substring of that text. 
Better to see something than nothing, right?


avently:
@menny do you want this changes?  I'll remove mHintOverflowLabel from everywhere if you are agree (this will fix tests).

menny:
@avently yes, that's fine: you can remove `mHintOverflowLabel`

avently:
@menny done. What is codecov wanting from me?

menny:
Don't worry about that.

menny:
Thanks

PR #1187, opened 2017-12-07
Added top and bottom row style

avently:
I added simple style for bottom and top row. It is based on ext_kbd_top_row_numbers and ext_kbd_bottom_row_alt.
I just removed unused symbols from ext_kbd_top_row_numbers. 
Also I removed smile and added characters selection button to ext_kbd_bottom_row_alt and some symbols. That's it.

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -344,6 +344,7 @@
   <string name="extension_kbd_top_small_numbers">Цифры</string>
   <string name="extension_kbd_top_numbers_alt">Альтернативные цифровые кнопки</string>
   <string name="extension_kbd_top_nav">Клавиши навигации</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values-ru/strings.xml
```diff
@@ -344,13 +344,15 @@
   <string name="extension_kbd_top_small_numbers">Цифры</string>
   <string name="extension_kbd_top_numbers_alt">Альтернативные цифровые кнопки</string>
   <string name="extension_kbd_top_nav">Клавиши навигации</string>
+  <string name="extension_kbd_top_numbers_simple">Без лишнего, маленькие цифры</string>
```
menny: No need to localize here, use https://crowdin.com/project/anysoftkeyboard for that.

PR #1194, opened 2017-12-15
Popup text animation control

menny:


menny:
Set pref value when service restarts
Do not pop when manually picking suggestion

PR #1294, opened 2018-03-23
Switch capitalisation of highlighted word

ViXXoR:
Pressing the shift key with characters highlighted toggles the characters from uppercase to lowercase, based on whether or not they are currently uppercase or lowercase.

I'm not sure if there should be a preferences toggle to enabled/disable this. If there should be, then that's something I can look into adding.


menny:
this is great. Thank you.

Two comments: 
1. Calling `getSelectedText` is an IPC call, can you figure away to avoid calling this, or avoiding calling that IPC method? One way could be to leverage `onUpdateSelection` method
2. We Ned unit tests for this

ViXXoR:
I've made changes based on your comments. Let me know if there is anything else this needs.

Issue #1278, opened 2018-03-03
Keyboard Does Not Close After App Closes, Keyboard Stays Open On Home screen


shagele:
### Steps to reproduce
1.Open any app that uses the keyboard.
2.Type some information into the app.
3.Leave keyboard open in the app but close the app.

### Expected behaviour
Keyboard should close and leave a normal looking home screen.

### Actual behaviour
Keyboard remains open on the home screen.

**Nougat 7.0

**Device manufacturer and model:**Samsung Galaxy S8

**List of installed add-ons (like languages, or themes):**English language, Nova launcher

menny:
I noticed it is easy to reproduce if I open the Emoji or utility keyboard before hitting the HOME key.

ViXXoR:
So I've been looking into this, and from what I can tell there's no way from the keyboard to know how the user is exiting the keyboard view. I tried different things like using an EditorInfo to try and see if there is a valid input, or trying to hook into onFinishInput() or various other functions.

I have some working code that can solve the issue, but what it does is if the user pushes back from the Emoji keyboard, the keyboard is hidden. If the user wants to go back to the default keyboard, they need to hit the key on the emoji keyboard that takes them back.

I played another with other keyboards, and it seems they all have this behaviour, and I assume it's due to the issues we are seeing here.

I'll open a PR to show the small fix, in case there is interest.

PR #1298, opened 2018-03-27
Fix Keyboard not Closing when Emojis open

ViXXoR:
Fix for problem described in #1278

This changes the behaviour of the keyboard slightly when the emoji keyboard is up and hardware back/home buttons are pressed.

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
+++ b/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
@@ -1093,6 +1093,9 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardWithGestureTyping {
                 break;
             case KeyCodes.CANCEL:
                 mCancelKeyPressed = true;
+                if(mUtilityKeyboardShown) {
+                    mUtilityKeyboardShown = false;
+                }
                 hideWindow();
                 break;
             case KeyCodes.SETTINGS:
@@ -1124,6 +1127,7 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardWithGestureTyping {
                 }
                 break;
             case KeyCodes.UTILITY_KEYBOARD:
+                mUtilityKeyboardShown = true;
                 getInputView().openUtilityKeyboard();
                 break;
             case KeyCodes.MODE_ALPHABET_POPUP:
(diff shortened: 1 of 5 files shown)


menny:
I have a theory about this issue: I think that we are not handling `onFinishInput()` correctly. I think that this method is called when the OS does not need the IME anymore, but we (as in, AnySoftKeyboard) fail to completely hide the window.
I'm not sure if calling `closing()` is the right thing to do there or the only thing.

ViXXoR:
Based on the life cycle of an IME, I believe that the keyboard is acting as
expected. The main issue is that when the emoji keyboard is open (another
instance of a class inheriting from AnySoftKeyboardBase), and a request to
close comes in, the emoji keyboard will remove itself, and display the
standard view again.
Unfortunately, from what I understand in my testing, there is no way to
determine where the request to hide the emoji keyboard is coming from
(unless it originates from the keyboard itself). So it may be the user
pressing a hardware back key, or the hardware home key. In both cases the
behaviour will be the same.

I may not have enough of an understanding at the moment, so I will spend
some more time to see if maybe I'm missing something.

menny:
Thank you

ViXXoR:
So I've been testing and examining some more, including browsing through the AOSP code for the LatinIME for some potential insight.

It seems that the `onFinishInput()` and `onFinishInputView()` functions are doing what they are expected to be doing, which is to clean up the view(s) before closing and potentially moving on to the next input field.

I still believe the issue is within the `hideWindow()` and `handleCloseRequest()` functions. When those functions are called, the expectation is that the keyboard will close/disappear. Within `AnySoftKeyboardWithQuickText` the `handleCloseRequest()` function calls `cleanUpQuickTextKeyboard(...)`, which hides the Emoji/QuickText view if it is visible, but then shows the standard ASK view. I understand that this functionality is aimed at allowing a user to insert emojis and then hit back to continue typing. However, this is a problem as I was unable to find any way to detect the reason that the keyboard is closing, be it hardware buttons (home or back) or change of focus. So no matter what, if the Emoji/QuickText keyboard is up, when any hardware button is pushed, only the Emoji/QuickText keyboard is hidden.

At this point I'm not sure what any alternative fixes might be. As I mentioned before, all other keyboards I've been looking at and playing with completely close when the back hardware button is pressed, even if an alternate view is up at the time of the back button being pressed. I'm thinking that this is due to the limitation of `hideWindow()` not knowing the origin of the request.

menny:
Okay, got it. Good findings.
It sounds like we can still distinguish between whether the close request was made because of BACK, close-key-pressed or directly to hideWindow. No?

ViXXoR:
I think you're correct. I do know that the HOME button isn't captured by
the keyboard, but I think BACK may be.

menny:
@ViXXoR I see this: com/anysoftkeyboard/ime/AnySoftKeyboardHardware.java:102.

ViXXoR:
I've got a working solution to the Emoji/QuickText and utility keyboard issues. Curious what you think about it.

menny:
thanks!

Also, great unit-tests.

PR #1303, opened 2018-03-28
Fix spacing delete and shift for various combinations of typing and gesturing

justinmichaud:
- Fix spacing and delete for various combinations of typing and gesturing
- Fix spacing and delete for input types (ex: chromium url bar) that don't do auto space
- Delete whole word when deleting a gesture
- Fix shift, caps lock and finishing a gesture by pressing shift

I undid your TextState changes. This implementation has the advantage of working correctly for the cases above (especially the chromium url bar), but it is definitely an abuse of the states. I am not sure how else to implement it though, so I am definitely open to ideas. The tests should handle any regressions if you want to play around with it.

Everything related to gesturing should be inside the withGestureTyping class, which should hopefully make it less awful than the last PR. Also, everything has a check, so this should not introduce any regressions if the feature flag is disabled.

Thanks for getting the mocks set up! They were so helpful. Let me know if you want this PR broken up.


menny:
I've implement this in a different PR

PR #1308, opened 2018-03-30
Load gesture typing words list from jni

justinmichaud:
This seems to be working properly in my casual testing. I am open to suggestions if you have a cleaner way of doing this PR, but it should be good to go.

This PR loads words from the built in dictionary. It also adds support for accents to the gesture detector. I have tested it with both english and french (although the size of the french dictionary makes gesturing impractical; even simple words like été have too many similar words close by, making the low accuracy of the gesture detector painfully obvious)

To allow gesturing after switching keyboards, I updated the Rx code to allow it to be cancelled, although I think I am missing something because it is a mess now. One bug that I noticed (both before and after this patch) is that spamming the language button crashes the keyboard. I don't think it is in the scope of this PR to fix, but once this is merged I will see about looking into it.


bocekm:
@justinmichaud, I have installed the debug build from the CircleCI (https://circleci.com/gh/AnySoftKeyboard/AnySoftKeyboard/2667#artifacts/containers/0). I was already writing a comment here that I'm not getting any results at all. But it probably just takes some time to build the dictionary or network or whatever you use under the hood. I switched on two dicts (en+cs) and it took about one minute to process. It now really uses both dicts and it's really good in picking the right words. Thanks a lot and I'm looking forward to have it in the main build.

bocekm:
Could you make the built network/db persistent? When I leave the phone be for some time, I guess the Android optimization kicks in and kills the keyboard process, and then when I want to start typing again, it takes about 15 seconds before the swiping feature is ready.

ildar:
@justinmichaud , devs,
are there plans to go on with this one?

justinmichaud:
@ildar I am going to be done exams in ~2 weeks, so I can take another look

ildar:
Thanks a lot!
Good luck with it!

Issue #998, opened 2017-05-27
Reversed backspace and shift keys on Colemak layout


ghost:
### Steps to reproduce
1. Download alternative English layouts 
2. Set Colemak as layout 
3. No option to.set backspace on right 

### Expected behaviour
See above 

### Actual behaviour
See above 

**Android OS version:**
7.0
**Device manufacturer and model:**
Motorola Moto G5+
**List of installed add-ons (like languages, or themes):**
Hebrew
Spanish
Alternative English layouts

PR #1355, opened 2018-05-25
Fix swapped shift and backspace on colemak layout

mesilliac:
ASK's Colemak layout swaps shift and backspace, as noted in issue #998 . This swapped layout does not match the backspace key positioning on the number and symbol input pages, leading to much annoyance.

This PR simply swaps them back to match the symbols page, and all other keyboards.

~The official colemak keyboard layout (colemak.com) does not move the backspace key at all, it is in the default position on the keyboard.~ Actually colemak recommends remapping capslock to backspace, but on an android keyboard this doesn't make much sense.


menny:
Thank you

PR #1416, opened 2018-07-08
Create currency_data.xml for uk

burunduk:
Add «HRYVNIA SIGN»


menny:
Thank you. Do you know since which Android version was this symbol supported?

burunduk:
The sign was Unicode encoded as HRYVNIA SIGN in version 4.1, as published in 2005.
I no information about the android version.

Issue #1443, opened 2018-08-26
[Feature Request]  Terminal Common Top Row


returntrip:
### Current behaviour
Currently we can use the "Tetminal" keyboard to access "Ctrl", "Esc", etc. This is great but this layout shifts the letters 1 row up and, for people typing by memory, the wrong keys are being hit when typing (pressing "Esc" when wanting to press backspace, "Tab" instead of "z", etc). This is.really annoying. 

### Suggested behaviour
Add the terminal keyboard "special keys" (Ctrl, tab, arrows, Esc. etc etc) to the "Common Top Generic Row" in UI

Hope you think it is a good.idea.

PR #1455, opened 2018-09-09
Add terminal common top row

ajgrf:
Adds a terminal common top row, for terminal use with non-QWERTY layouts, or for people who prefer these keys on top.

See #1443.


menny:
Screenshot?

ajgrf:
screenshot_termux_20180908-193614

Issue #111, opened 2013-04-02
Add option to change layout-switch shortcut


AlexCzar:
Many people use bluetooth keyboards with their android devices (as do I). Some of us need more than one layout and ASK is the best keyboard for this (in my opinion at least). Current shortcut (shift+space) to change layout is counterintuitive for any person used to hardware keyboards. Can you please add the option to select another shortcut (preferably customizable) e.g. on my PC I use Shift+CapsLock (which also isn't a standard, I know), but at least addition of Ctrl+Shift and Alt+Shift options is a must.

ildar:
I'd also suggest a combination without Shift: the KeyRelease for Shift may come too late and pressing SPACE produces unintended layout switch.
Example: Typing the text "Choose the best OS now!" makes the layout switch before "now" word, because the "S" (with Shift+s) followed by the SPACE so ASK switched the layout :(((
Please fix. Or should I file another bug for this?

PR #1465, opened 2018-09-20
Add toggle for physical hardware shift+space layout switch shortcut

antonok-edm:
Issue #111 mentioned that the current shift+space shortcut for switching layouts on physical keyboards was unintuitive, and it is inconvenient when typing capital letters, parentheses, underscores, etc.

This PR adds a toggle in the hardware keyboard settings to disable that shortcut.

### DIFF
--- a/app/src/main/res/values-en/strings.xml
+++ b/app/src/main/res/values-en/strings.xml
@@ -174,12 +174,6 @@
     <string name="hide_soft_when_physical_off_summary">Show soft keyboard when physical keyboard is
         used
     </string>
-    <string name="enable_alt_space_language_shortcut">Alt + Space switches languages</string>
-    <string name="enable_alt_space_language_shortcut_on_summary">Alt + space switches to the next available layout.</string>
-    <string name="enable_alt_space_language_shortcut_off_summary">Alt + space enters a space.</string>
-    <string name="enable_shift_space_language_shortcut">Shift + Space switches languages</string>
-    <string name="enable_shift_space_language_shortcut_on_summary">Shift + space switches to the next available layout.</string>
-    <string name="enable_shift_space_language_shortcut_off_summary">Shift + space enters a space.</string>
     <string name="fullscreen_input_connection_supported">Use full-screen editor in landscape
     </string>
     <string name="fullscreen_input_connection_supported_on_summary">Take over the screen when
--- a/app/src/main/res/values-en/strings.xml
+++ b/app/src/main/res/values-en/strings.xml
@@ -174,6 +174,9 @@
     <string name="hide_soft_when_physical_off_summary">Show soft keyboard when physical keyboard is
         used
     </string>
+    <string name="enable_shift_space_language_shortcut">Shift + Space switches languages</string>
+    <string name="enable_shift_space_language_shortcut_on_summary">Shift + space switches to the next available layout.</string>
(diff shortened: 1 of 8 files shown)


shaggy16:
Add

menny:
Thanks.
Can you add another settings to control ALT+SHIFT too?

antonok-edm:
@menny I can certainly do that, but I'd like to have a little more background information before I make any changes there. I opened #1467 for further discussion.

I also propose completing that in a separate PR, since the functionality seems to be independent of the Shift+Space shortcut.

app/src/main/res/values-en/strings.xml
```diff
@@ -174,6 +174,12 @@
     <string name="hide_soft_when_physical_off_summary">Show soft keyboard when physical keyboard is
         used
     </string>
+    <string name="enable_alt_space_language_shortcut">Alt + Space switches languages</string>
```
menny: You don't need to specify the strings for `values-en/strings.xml`, crowd-in service generates those automatically (directly copies from `values/strings.xml`).

app/src/main/res/values/strings.xml
```diff
@@ -174,6 +174,12 @@
     <string name="hide_soft_when_physical_off_summary">Show soft keyboard when physical keyboard is
         used
     </string>
+    <string name="enable_alt_space_language_shortcut">Alt + Space switches languages</string>
```
menny: I believe we should use uppercase _SPACE_, and not _Space_. What do you think?

app/src/main/res/values/strings.xml
```diff
@@ -174,6 +174,12 @@
...
+    <string name="enable_alt_space_language_shortcut_on_summary">Alt + space switches to the next available layout.</string>
```
menny: Maybe indicate that this is affecting only physical keyboards?
`Hardware ALT + SPACE switches to the next available layout.`

app/src/main/res/values/strings.xml
antonok-edm: I was trying to be consistent with the style of some of the other config options -- e.g. "Space switches keyboards": "Switch from symbols/numbers to alphabet when pressing space".
I've found a mix of "Space" and "space", but I don't see any "SPACE".

app/src/main/res/values/strings.xml
antonok-edm: It is in the "Physical Keyboard Tweaks" section, but the headers are fairly small so I think that makes sense.

app/src/main/res/values/strings.xml
menny: Then capitalizrd `Space`. It's a name of  key.

antonok-edm:
@menny Should be all set.

menny:
Thanks!

poet-man:
Not being familiar with it … how/where can I switch off that shift-space for changing layouts?
I happen to activate it accidentally very often.
Thanks

antonok-edm:
@poet-man The option is called 'Shift + Space switches languages'; you can find it under Language > Tweaks and More > Hardware Keys Mapping > Tweaks > Physical Keyboard Tweaks

poet-man:
Thanks a lot

Am 6. November 2018 22:03:31 schrieb Anton Lazarev <notifications@github.com>:

PR #1596, opened 2019-01-23
Suggestions strip

menny:


mho-games:
I don't know where to put this...
Features Request: large charter (font) or adjustable font for keyboard.  Great produce I am glad to be a supporter of your work.
But I hope that there will a way to adjust font size... think its same keybord same font on a 9" nexus and postage stamp LGE smart phone    .. your simplest keyboard on with large print to cure fat fingers "or people like me in a clean room all dressed up".. fewer buttons easier to read would a game changer

Issue #1610, opened 2019-01-30
Remove PhoneKeyboard from listed keyboards


VitaliPom:
When using symbols and typing with Markdown format for example, this useful but annoying feature in this keyboard will kill you shortly. Remove this view here, it will pop when you need to use numbers only in web fields for example. Let the view switch toggle only for special keys and symbols keyboard.

VitaliPom:
Priority 1
Kill!

VitaliPom:
CODE 
Dexamol

VitaliPom:
FIXED 
Waiting to be pushed (Looking forward to be added to the list of Contributors for this)

PR #1625, opened 2019-02-06
FIX *Remove PhoneKeyboard from listed keyboards #1610*

VitaliPom:
Now typing is a little bit more easy :)

DESCRIPTION "Keyboard just turned out symbols to be more intuitive and
now symbols will not include the irrelevant Numpad keyboard. It will pop
up only when keyboard for num pad is set up in a web app mobile view.
So if you need to type up numbers, you will have it. For everything else you have full support for symbols from now on."


VitaliPom:
@menny

VitaliPom:
https://youtu.be/QUJ71GWWV0Y

VitaliPom:
@menny confirm this one as well please. It's also a quick fix to a common problem...

menny:
I'm closing this. It is way back. If you want to try again, open a new PR

PR #1665, opened 2019-03-17
Repaint algorithm

VitaliPom:
preview, non complete version.


VitaliPom:
hi @menny got a progress on that, will update in future when something is ready, still some bugs.  Minor though, unlike in the past.

PR #1694, opened 2019-04-12
Gesture typing improvements

PFischbeck:
This pull request greatly improves the speed+quality of gesture typing, at least for me!
In particular:

- **Faster path generation**: From 12s to 3.5s (English, debug mode), from 17s to 3s (German, debug mode)
- **Faster candidate finding**: From 120ms to 80ms when typing "implementation" on an English Keyboard (in debug mode)
- **Better special character handling**: Previously, when a keyboard had two keys with the same base character (e.g., "ä" and "a" on the German keyboard), the path generation would always choose the base character, making it nearly impossible to gesture type words like "Männer". This now works properly.
- Most importantly: **Much better word recognition**: At least for me, the recognition of the gesture typed words is much better, especially for long words. For this, I mainly reduced the number of points that are chosen for the path gesture-typed by the user. The points on this path are now almost only the ones where the path changes direction, leading to a higher accuracy.

Some more details:
- I split up the variable `mCurvatureSize`. Previously, it was used with two meanings: as the minimum distance between two points, and as the number of points to the left/right used for curvature calculation. We now have `CURVATURE_NEIGHBORHOOD` and `mMinPointDistance`.
- Most of the speedup also comes from a `mKeysByCharacter` lookup, which is now used. Previously, we iterated over all keys to check whether they matched the character we're looking for.
- I added some more debug logging for easier measuring of the performance.

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
+++ b/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
@@ -27,8 +27,8 @@ public class GestureTypingDetector {
     private static final double CURVATURE_THRESHOLD = Math.toRadians(170);
     // How many points away from the current point do we use when calculating curvature?
     private static final int CURVATURE_NEIGHBORHOOD = 1;
-    // How far away do two points of the gesture have to be?
-    private final int mMinPointDistance;
+    // How far away do two points of the gesture have to be (distance squared)?
+    private final int mMinPointDistanceSquared;
 
     private final ArrayList<CharSequence> mCandidates = new ArrayList<>(64);
 
@@ -58,7 +58,7 @@ public class GestureTypingDetector {
     private final ArrayList<int[]> mWordsCorners = new ArrayList<>();
 
     public GestureTypingDetector(int minPointDistance, @NonNull Iterable<Keyboard.Key> keys) {
-        mMinPointDistance = minPointDistance;
+        mMinPointDistanceSquared = minPointDistance * minPointDistance;
         mKeys = keys;
 
         mGenerateStateSubject.onNext(LoadingState.NOT_LOADED);
@@ -166,7 +166,7 @@ public class GestureTypingDetector {
             final int dx = mWorkspaceData.mCurrentGestureXs[mWorkspaceData.mCurrentGestureArraySize - 1] - x;
             final int dy = mWorkspaceData.mCurrentGestureYs[mWorkspaceData.mCurrentGestureArraySize - 1] - y;
 
-            if (dx * dx + dy * dy <= mMinPointDistance * mMinPointDistance) return;
+            if (dx * dx + dy * dy <= mMinPointDistanceSquared) return;
         }
 
         mWorkspaceData.addPoint(x, y);
--- a/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
+++ b/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
@@ -24,7 +24,7 @@ import io.reactivex.subjects.ReplaySubject;
 public class GestureTypingDetector {
     private static final String TAG = "GestureTypingDetector";
 
-    private static final double CURVATURE_THRESHOLD = Math.toRadians(160);
+    private static final double CURVATURE_THRESHOLD = Math.toRadians(170);
     // How many points away from the current point do we use when calculating curvature?
     private static final int CURVATURE_NEIGHBORHOOD = 1;
     // How far away do two points of the gesture have to be?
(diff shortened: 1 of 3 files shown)


menny:
this is insane! Thank you.

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -23,9 +24,11 @@
...
     private static final String TAG = "GestureTypingDetector";
 
-    private static final double CURVATURE_THRESHOLD = Math.toRadians(160);
-    // How many points away from the current point to we use when calculating curvature?
-    private final int mCurvatureSize;
+    private static final double CURVATURE_THRESHOLD = Math.toRadians(170);
+    // How many points away from the current point do we use when calculating curvature?
+    private static final int CURVATURE_NEIGHBORHOOD = 1;
+    // How far away do two points of the gesture have to be?
+    private final int mMinPointDistance;
```
menny: This variable is only used with its square value, do you think we can just store the square value?

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -54,10 +57,17 @@
     private final ReplaySubject<LoadingState> mGenerateStateSubject = ReplaySubject.createWithSize(1);
     private final ArrayList<int[]> mWordsCorners = new ArrayList<>();
 
-    public GestureTypingDetector(int curvatureSize, @NonNull Iterable<Keyboard.Key> keys) {
-        mCurvatureSize = curvatureSize;
+    public GestureTypingDetector(int minPointDistance, @NonNull Iterable<Keyboard.Key> keys) {
+        mMinPointDistance = minPointDistance;
         mKeys = keys;
 
+        for (Keyboard.Key key : mKeys) {
```
menny: Can this be moved into the `generateCornersInBackground` method? So the calculation would be done in a background thread and not on the UI thread (which calls this constructor)

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -117,37 +127,33 @@ public void destroy() {
...
-    private static int[] generatePath(char[] word, Iterable<Keyboard.Key> keysList, int curvatureSize, WorkspaceData workspaceData) {
+    private static int[] generatePath(char[] word, SparseArray<Keyboard.Key> keysByCharacter, WorkspaceData workspaceData) {
         workspaceData.reset();
         //word = Normalizer.normalize(word, Normalizer.Form.NFD);
         char lastLetter = '\0';
 
         // Add points for each key
         for (char c : word) {
-            c = Dictionary.toLowerCase(c);
+            c = Character.toLowerCase(c);
```
menny: So, you're saying it's better to use `Character.toLowerCase`? Should we replace all `Dictionary.toLowerCase` calls to that?

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -117,37 +127,33 @@ public void destroy() {
...
-                    }
-                }
-            }
+            Keyboard.Key keyHit = keysByCharacter.get(c);
 
             if (keyHit == null) {
-                Logger.w(TAG, "Key %s not found on keyboard!", c);
-                continue;
+                // Try finding the base character instead, e.g., the "e" key instead of "é"
+                char baseCharacter = Dictionary.toLowerCase(c);
```
menny: I see that here you used `Dictionary. toLowerCase` instead of the `Character` counterpart, why?

PFischbeck: `Character.toLowerCase` gives you the actual lowercase character, e.g. `É` --> `é`. But `Dictionary.toLowerCase` gives you the lowercase base character, e.g., `É` --> `e`. Previously, the gesture typing was always using `Dictionary.toLowerCase`, which meant there was no difference between, say, `ä` and `a`. However, for keyboards, where both characters are present as different keys (e.g., the German keyboard), we want a different behavior in the path generation.
If a key for `ä` is present on the keyboard, we assume the user would swipe over it to use it, and not the character `a`. This allows for him to type words like `Männer` properly. Therefore we look for the `Character.toLowerCase`.
But if they want to type `café` on a German keyboard, there is no key for `é`. Instead, we would assume they would swipe over `e`. Therefore we look for the `Dictionary.toLowerCase` if we couldn't find the `Character.toLowerCase`.
This is why we should first look for a key which matches the lowercase exactly, and if we don't find such a key, we try looking for its base character.

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
PFischbeck: It depends, and I think we often would use both. See my other comment on the characters.

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
PFischbeck: Sure, I changed it so we store the square.

app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
PFischbeck: I moved it to a background thread. Please check whether this looks right, I'm pretty unfamiliar with RxJava...

menny:
thank you!

menny:
Can I mention you in the changelog? In regard to this improvement?

PFischbeck:
Sure, go ahead!

topas-rec:
I just tested it by typing this comment.  IT IS AMAZING! Many thanks!

PR #1886, opened 2019-11-21
Allow displaying hints with the AOSP based themes

### LINKED ISSUE #1884: Add hints to AOSP theme keyboards
Hi, this would just be a case of removing the hintTextSize 0 item in styles_aosp_keyboard_theme, and would allow showing hints in these themes.

Currently on mobile so can't submit a PR but will do tonight if no-one sneaks it in before hand

RichyHBM:
Fixes #1884 by allowing the user to chose if they want to display hints or not with AOSP themes


menny:
Thanks

Issue #620, opened 2016-04-06
Cursor acts as if it's in another place


ashirviskas:
### Steps to reproduce
1. Enable multitap special characters in other languages and word suggestions
2. Type a few fords
3. Go back those few words and make a double space
4. Type a word inbetween double space, the word must end with a special character, typed by multitaping (I.e. „Eglė” or „Liepą”)
5. Tap in the end of the string to place cursor in there
6. Press backspace
   ### Expected behaviour
   The last letter of the string should be deleted
### Actual behaviour

Cursor appears in the last place it's been — after the last special character (I.E. after Liepą) and then deletes the letter in a middle of this string.

I'm using Lithuanian language pack

menny:
Can you try out the latest beta release?
Sign up here: https://play.google.com/apps/testing/com.menny.android.anysoftkeyboard
and download from the Play-Store

ashirviskas:
I tried, same.

Issue #303, opened 2014-01-18
(Restore Functionality): Show current language in the status bar


menny:

menny:
Fixed a while ago

PR #1887, opened 2019-11-24
Generate unsigned release APKs, fix #971 and #1636

### LINKED ISSUE #971: Return Key is Drawn off the Screen
The green return key is drawn on my screen with a right margin too far to the right, a little off the screen. It doesn't line up with the keys above it.

I have a Nexus 5X with Android 6.0.1 and I am using "English QWERTY Latin keyboard with symbols (on long press)" and the "New Generation - Testing" layout.

Does anyone else see this?

### LINKED ISSUE #1290: Cursor jumps back automatically
### Steps to reproduce
1. Type quickly 
2.
3.

### Expected behaviour
Tell us what should happen

Typed symbols should appear in the order they were typed

### Actual behaviour
Tell us what happens instead

Cursor lags behind or jumps back and you start typing in the middle of the word. For example if I type quickly the word "example", I get something like "expleam"

**Android OS version:**

8.0

Huawei mate 9

Lithuanian language and main AnySoftKeyboard app.

### LINKED ISSUE #1636: Pressing backspace on special characters causes an unknown character to appear
### Steps to reproduce
1. Type any special character with a Unicode value above U+FFFF. For example, emoji such as 😃, or double-struck math letters (which is how I found this bug) such as 𝔾.
2. Press space
3. Press backspace

### Expected behaviour
Pressing delete after a space should only delete the space.

### Actual behaviour
The space is deleted and the special character is replaced with �.

**Android OS version:**
8.1.0, running Lineage OS

Google Nexus 6P

Only a custom language pack that I'm working on https://github.com/TheVirtualMachine/LanguagePack

This is a language pack for math symbols. I discovered this bug while testing the double-struck characters. This bug occurs for all double-struck characters except ℚ, ℝ, ℙ, ℍ, ℤ, ℂ, ℕ, which are precisely the double-struck characters with Unicode values BELOW U+FFFF.

ArenaL5:
These commits achieve four goals:

1. **Generate unsigned release APKs instead of debug APKs, with `./gradlew assembleRelease`.** Like you asked for at #303, the Gradle scripts for this projecy will now default to unsigned release APKs if no suitable keystore can be found. Debug APKs can still be created with `assembleDebug` and `build`. (I think `build` isn't working right now because of some problem with canary and test releases; I haven't looked into that.)

2. **Fix #1636 and all related bugs.** Basically, most string manipulation functions in Java depend in `char` values, which can each hold a single multi-byte character... until U+FFFF. Some characters have now assigned code-points over U+FFFF, and those can only fit in two `char`s.

Some select functions work with `int` values, which can fit values over 0xFFFF, and thus correspond to actual characters like we understand them. What I did here is to change every line which assumed "1 int = 1 char" and rigged the cursor keys to never split a character. **[This also allows keyboards to use codepoints over U+FFFF in `popupCharacters`.](https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/1636#issuecomment-466153250)**

I also fixed some very situational glitches (like deleting at prediction boundaries, and checking for the length of a word). This ~*should* fix bug #620 and~ *may* fix some occurrences of #1290, but not the ones that come from a race condition. (See below)

I needed to extend your WordComposer API to account for that. `length()` now returns the number of actual characters (codePoints), a new function `charLength` returns the number of char values, and `deleteLast` and `deleteForward` now return the number of `char` values deleted.

3. Fix some glitches when reporting to the logcat. **I may have overstepped my bounds here and changed some `TAGs` on commit 709ed5bb5854cd825eaf048be0ac784dedf7dff3**. Let me know if you'd like to undo these name changes or to use something else.

4. **Fix #971**, a cosmetic issue. The positioning of every key after a hidden key (with `ask:showInLayout="if_applicable"` was some pixels off.

What this PR won't fix
-----

- As described above, it's going to be difficult to fix #1290 if it's caused by race conditions.
- I misremembered bug #620, it's a distinct bug and still not fixed. (Multitap confuses the cursor position.) If this goes well, I'll try to fix it in the near future.
- Selection still acts a bit wonky: both left and right keys extend the selection range, none of them shrink it, and pressing up or down will abort it. Would you like a system like Shift+←↑→↓ on a PC, maybe with a button to swap regular cursor and selection cursor?

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
+++ b/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
@@ -123,9 +123,10 @@ public abstract class AnySoftKeyboardClipboard extends AnySoftKeyboardSwipeListe
                         ic.setSelection(globalSelectionStartPosition, globalCursorPosition);
                     } else {
                         ic.setSelection(
-                                globalSelectionStartPosition -
-                                    Character.charCount(toLeft.codePointBefore(toLeft.length())),
-                                    globalCursorPosition);
+                                globalSelectionStartPosition
+                                        - Character.charCount(
+                                                toLeft.codePointBefore(toLeft.length())),
+                                globalCursorPosition);
                     }
                     return true;
                 case KeyEvent.KEYCODE_DPAD_RIGHT:
@@ -135,8 +136,7 @@ public abstract class AnySoftKeyboardClipboard extends AnySoftKeyboardSwipeListe
                     } else {
                         ic.setSelection(
                                 globalSelectionStartPosition,
-                                globalCursorPosition +
-                                    Character.charCount(toRight.codePointAt(0)));
+                                globalCursorPosition + Character.charCount(toRight.codePointAt(0)));
                     }
                     return true;
                 default:
--- a/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
+++ b/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
@@ -116,11 +116,28 @@ public abstract class AnySoftKeyboardClipboard extends AnySoftKeyboardSwipeListe
         if (mArrowSelectionState && ic != null) {
             switch (keyEventKeyCode) {
                 case KeyEvent.KEYCODE_DPAD_LEFT:
-                    ic.setSelection(
-                            Math.max(0, globalSelectionStartPosition - 1), globalCursorPosition);
+                    // A Unicode codepoint can be made up of two Java chars.
+                    // We check if that's what happening before the cursor:
+                    final String toLeft = ic.getTextBeforeCursor(8, 0).toString();
+                    if (toLeft.length() == 0) {
+                        ic.setSelection(globalSelectionStartPosition, globalCursorPosition);
+                    } else {
+                        ic.setSelection(
+                                globalSelectionStartPosition -
+                                    Character.charCount(toLeft.codePointBefore(toLeft.length())),
+                                    globalCursorPosition);
+                    }
                     return true;
                 case KeyEvent.KEYCODE_DPAD_RIGHT:
-                    ic.setSelection(globalSelectionStartPosition, globalCursorPosition + 1);
+                    final String toRight = ic.getTextAfterCursor(8, 0).toString();
+                    if (toRight.length() == 0) {
+                        ic.setSelection(globalSelectionStartPosition, globalCursorPosition);
+                    } else {
+                        ic.setSelection(
+                                globalSelectionStartPosition,
+                                globalCursorPosition +
+                                    Character.charCount(toRight.codePointAt(0)));
+                    }
                     return true;
                 default:
                     mArrowSelectionState = false;
--- a/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
+++ b/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
@@ -17,6 +17,7 @@ public abstract class AnySoftKeyboardClipboard extends AnySoftKeyboardSwipeListe
 
     private boolean mArrowSelectionState;
     private Clipboard mClipboard;
+    protected static final int MAX_CHARS_PER_CODEPOINT = 2;
 
     @Override
     public void onCreate() {
@@ -118,7 +119,8 @@ public abstract class AnySoftKeyboardClipboard extends AnySoftKeyboardSwipeListe
                 case KeyEvent.KEYCODE_DPAD_LEFT:
                     // A Unicode codepoint can be made up of two Java chars.
                     // We check if that's what happening before the cursor:
-                    final String toLeft = ic.getTextBeforeCursor(8, 0).toString();
+                    final String toLeft =
+                            ic.getTextBeforeCursor(MAX_CHARS_PER_CODEPOINT, 0).toString();
                     if (toLeft.length() == 0) {
                         ic.setSelection(globalSelectionStartPosition, globalCursorPosition);
                     } else {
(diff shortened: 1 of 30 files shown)


ArenaL5:
There's a problem with the code, I hadn't noticed in my local machine. I'll fix it today.

ArenaL5:
The code should work. I debugged it extensively before commit 20c695bc7b73653747c2f74731a8b0c36489b52e and done some quick tests right now.

Codecov throws an error, which is weird because there should be more comments now. (Or I'm misunderstanding the purpose of Codecov, both are possible.)

dictionaries/src/main/java/com/anysoftkeyboard/dictionaries/WordComposer.java
```diff
@@ -63,12 +63,21 @@ public void reset() {
     }
 
     /**
-     * Number of keystrokes in the composing word.
+     * Number of keystrokes (codepoints, not chars) in the composing word.
      *
      * @return the number of keystrokes
      */
     @Override
     public int length() {
```
menny: I guess a better name for this is `codesCount`?

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
```diff
@@ -116,11 +116,28 @@ protected boolean handleSelectionExpending(
         if (mArrowSelectionState && ic != null) {
             switch (keyEventKeyCode) {
                 case KeyEvent.KEYCODE_DPAD_LEFT:
-                    ic.setSelection(
-                            Math.max(0, globalSelectionStartPosition - 1), globalCursorPosition);
+                    // A Unicode codepoint can be made up of two Java chars.
+                    // We check if that's what happening before the cursor:
+                    final String toLeft = ic.getTextBeforeCursor(8, 0).toString();
```
menny: Why `8`? Is this the longest possible multi-codepoints?
In any case, can you extract that to a const? `protected static final int MAX_MULTI_CODEPOINTS_COUNT = 8;` and use that anywhere you used `8`?

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
```diff
@@ -709,16 +713,18 @@ protected void handleSeparator(int primaryCode) {
                 // input-box
                 // into "word "->"word, "
                 ic.deleteSurroundingText(1, 0);
-                ic.commitText(((char) primaryCode) + (newLine ? "" : " "), 1);
+                ic.commitText(new String(new int[] {primaryCode}, 0, 1) + (newLine ? "" : " "), 1);
```
menny: I see`new String(new int[] {primaryCode}, 0, 1)` is used. Can you extract that into a method and use that? Say `static String codePointToString(int code)` It will give a lot of context.

app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
```diff
@@ -272,11 +272,12 @@ private void initKeysMembers(Context askContext, KeyboardDimens keyboardDimens)
                 for (int keyIndex = rowStartIndex; keyIndex < rowEndIndex; keyIndex++) {
                     final Key keyToModify = keyList.get(keyIndex);
                     keyToModify.width = (int) (keyToModify.width + additionalSpacePerKey);
+                    keyToModify.x = (int) (keyToModify.x + xOffset);
                     if (keyIndex == foundLanguageKeyIndex) {
-                        xOffset -= (widthToRemove + keyboardDimens.getKeyHorizontalGap());
+                        xOffset -= widthToRemove;
```
menny: nice

dictionaries/src/main/java/com/anysoftkeyboard/dictionaries/WordComposer.java
ArenaL5: _Thank you._ That's a great idea, having `length()` meaning different things depending on the type of variable is very confusing.

I couldn't find a way remove completely the method `length()` or even print deprecation notices to the logcat, because it's also called on `KeyCodesProvider`-typed variables and... I'm not knowledgeable with Java enough to affect `WordComposer` without affecting `KeyCodesProvider`.

So I chose `codePointCount()` and `charCount()` (named after related functions at [String](https://developer.android.com/reference/java/lang/String.html#codePointCount(int,%20int)) and [Character](https://developer.android.com/reference/java/lang/Character.html#charCount(int))). `length()` is a deprecated synonym for `codePointCount()`, for the moment.

Commit 77f28dc9cbf2c6b62f1d685e656981549a93a389. I forgot to add a comment for future developers, I'll add it now.

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java
ArenaL5: 8 is a senseless safeguard. A codepoint can either fill up 1 or 2 `char`, and judging from my research, 2 will be the maximum for the foreseeable future.

(When they "run out" of codepoints they'll need to tweak UTF-16, and there's no telling on how would they do it. I also expect that at that point Java will add more methods to treat codepoints in an abstract way, without treating the text as UTF-16.)

Constant extracted and set to 2 on commit 71a8b2ff4b34ce01a4b53be6bf245bf5f7ed28db.

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
ArenaL5: I need to use that in several files. That's the standard code to cast a codepoint into a string; I couldn't find any other way to create a `CharSequence` from the codepoint.

If you can point me to a place where I can define this method just once for the whole project, I'll do it.

app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
ArenaL5: Thank you!

menny:
Keep up the good work. Thank you

Issue #1810, opened 2019-08-08
[Request] Completely Disable Personalized Suggestions


Z3ROGRAV1TYx:
**Can we please get the option to completely disable personalized input suggestions? (Without "removing suggestions").**

What I mean is, similar to in GBoard, you can disable personalized suggestions but keep suggestions enabled. 
Example: If I type 'l8r' often, it shows up on the suggestion strip. So my proposal is to be able to disable this completely, but keep it so as your typing, it still shows suggestions based off common words. So if I type 'lat', then "Later" will show up on that suggestion strip. You just won't be seeing that _PERSONALIZED_ 'l8r' for example.

I hope that was clear enough, sorry if it was confusing.

samantohermes:
Disable user-dictionary and next word suggestions

Z3ROGRAV1TYx:
@samantohermes yes but I want to keep next word suggestoons on. (This is possible on gboard). It looks at what your typing and can suggest, but doesnt personalize any results in the strip.

samantohermes:
So disable instead user-dictionary and auto dictionary

ArenaL5:
The options are there, but indeed they're broken. Thanks for your report, @ Z3ROGRAV1TYx.

There are workarounds to disable word learning for now:

1. Open AnySoftKeyboard from your app drawer, and when you see the welcome screen, click the Language button below (the icon is a globe).
2. Click Configure correction behavior and other dictionaries.
3. **This option may be broken:**

[] *User-dictionary*
Suggest user added words (double-click a typed suggestion to edit)

So if that's not working, continue this guide.
4. Scroll down, and look for this option:

*Auto dictionary*
Automatically learn new words.

Ideally, you should be able to set it to "Do not learn new words", but that option was only repaired on AnySoftKeyboard Dev. If you're not willing to install the developers version, set it to "After some usage", so the keyboard will only learn a word after 10 uses.

5. Scroll up again, click Word editor, and remove your custom words. Repeat this step periodically if "Do not learn new words" doesn't work for you.

PR #1893, opened 2019-12-02
Bug fixes for #620, #1810; corrections for #1887, #1636.

ArenaL5:
These commits fix small but basic bugs. Two of them are caused by my previous code... sorry :sweat_smile: 

- Dictionary auto-learning, which was supposed to be fixed with #1887, wasn't correctly fixed. Details are given in the summary for commit 8159890377c75a89d841aec62bfd93bc214dffa9.
- #620 is fixed with commit 60e6e42f73d16be59eafdb9766c5dc5bec6e773a, but please check the summary; in doing so, I removed a part of code added by @menny in e041360c5f2ebb9bf0c4663cd4e38cf1cf93e83a. **This doesn't seem to break anything but please double-check.**
- After fixing dictionary auto-learning, I wrote a quick fix for #1810 which appears to work, even after reboot. The preference `settings_key_use_user_dictionary` wasn't actually connected to anything.
- Your backword function (is that supposed to be a pun?) no longer splits Unicode characters as of commit 1617de1818f799daa0b8d8bb2c2b5733892a8058, fully realizing the fix for #1636. I also made it delete letters _and digits_ in a run to deal more easily with big numbers and hexadecimal strings.
- I also attempted to made code more legible with commit 74bee1d671a17c3d8c607bf4a73fed724cae6571 (using constants taken from LanguagePack, as I couldn't find them anywhere within the AnySoftKeyboard base repo) and added a bash script for easily debugging ASK from Linux.

I'd really like your input on this PR.

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
+++ b/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
@@ -613,9 +613,17 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
                 final int cursorPosition;
                 if (mWord.cursorPosition() != mWord.charCount()) {
                     // Cursor is not at the end of the word. I'll need to reposition
-                    cursorPosition =
-                            mGlobalCursorPosition
-                                    + Character.charCount(primaryCode); /*adding the new character*/
+                    /* The code for tracking the current position is split among several files and difficult to debug.
+                    This has been proven to work in every case: */
+                    if (multiTapIndex > 0) {
+                        final int previousKeyCode = key.getMultiTapCode(multiTapIndex - 1);
+                        cursorPosition =
+                                mGlobalCursorPosition
+                                        + Character.charCount(primaryCode)
+                                        - Character.charCount(previousKeyCode);
+                    } else {
+                        cursorPosition = mGlobalCursorPosition + Character.charCount(primaryCode);
+                    }
                     ic.beginBatchEdit();
                 } else {
                     cursorPosition = -1;
--- a/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
+++ b/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
@@ -49,9 +49,7 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
 
     private static final long CLOSE_DICTIONARIES_DELAY = 10 * ONE_FRAME_DELAY;
 
-    private static final long MAX_TIME_TO_EXPECT_SELECTION_UPDATE = 1500;
-    // a year ago.
-    private static final long NEVER_TIME_STAMP = -1L * 365L * 24L * 60L * 60L * 1000L;
+    private static final long NEVER_TIME_STAMP = -1L * 365L * 24L * 60L * 60L * 1000L; // a year ago.
 
     private static final int UNDO_COMMIT_NONE = -1;
     private static final int UNDO_COMMIT_WAITING_TO_RECORD_POSITION = -2;
@@ -131,7 +129,6 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
      */
     private int mUndoCommitCursorPosition = UNDO_COMMIT_NONE;
 
-    private long mExpectingSelectionUpdateBy = Long.MIN_VALUE;
     /*
      * is prediction needed for the current input connection
      */
@@ -458,14 +455,6 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
             mUndoCommitCursorPosition = newSelStart;
         }
 
-        final boolean isExpectedEvent = SystemClock.uptimeMillis() < mExpectingSelectionUpdateBy;
-        mExpectingSelectionUpdateBy = NEVER_TIME_STAMP;
-
-        if (isExpectedEvent) {
-            Logger.v(TAG, "onUpdateSelection: Expected event. Discarding.");
-            return;
-        }
-
         if (TextEntryState.willUndoCommitOnBackspace()
                 && mUndoCommitCursorPosition == oldSelStart
                 && mUndoCommitCursorPosition != newSelStart) {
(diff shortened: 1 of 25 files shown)


scripts/adb.sh
```diff
@@ -0,0 +1,38 @@
...
+    cd "$(dirname "$0")"
+    while [ ! -d ".git" ]
+    do
+        cd ..
+        if [[ "$(pwd)" == "/" ]]; then
+            echo -e "${error}Aborting.${nocolor} Not inside a Git project."
+            exit 1
+        fi
+    done
+    tags=`grep -R 'TAG = ".*"' app/src/main/java/com/* jnidictionaryv1/src/main/java/com/* jnidictionaryv2/src/main/java/com/*`
```
menny: Interesting strategy. I was expecting the script to retrieve the PID for ASK and grep on that, but your approach will survive service restarts.

scripts/adb.sh
```diff
@@ -0,0 +1,38 @@
...
+    # We can go back to our original folder now:
+    cd "$oldpath"
+    tags="$(echo $tags | sed -E 's![a-z/A-Z12]*\.java: (protected |private )?(static )?(final )?String [A-Z_]* = "([^\"]*)";!\4!g')"
+    comm="adb logcat $(echo "$tags " | sed "s/ /:$1 /g")*:S"
```
menny: Will this work with Linux and macOS?

app/src/main/java/com/anysoftkeyboard/keyboards/Keyboard.java
```diff
@@ -458,6 +458,13 @@ public Key(
             centerY = y + height / 2;
         }
 
+        public int getMultiTapCode(int tapCount) {
+            final int codesCount = getCodesCount();
+            if (codesCount == 0) return KeyCodes.SPACE; // space is good for nothing
```
menny: `0` could be an option here? Or maybe it doesn't matter.

app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
```diff
@@ -528,7 +527,7 @@ CharSequence getPreviewText(Key key) {
                     ? anyKey.label.toString().toUpperCase(Locale.getDefault())
                     : anyKey.label;
         } else {
-            int multiTapCode = getMultiTapCode(key);
+            int multiTapCode = key.getMultiTapCode(mTapCount);
             // The following line became necessary when we stopped casting multiTapCode to char
             if (multiTapCode < 32) {
```
menny: I don't recall what was the assumption here. Probably `printable code`?

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
```diff
@@ -458,14 +455,6 @@ public void onUpdateSelection(
             mUndoCommitCursorPosition = newSelStart;
         }
 
-        final boolean isExpectedEvent = SystemClock.uptimeMillis() < mExpectingSelectionUpdateBy;
```
menny: i'm not sure I understand the problem.

scripts/adb.sh
```diff
@@ -29,7 +30,8 @@ else
     # We can go back to our original folder now:
     cd "$oldpath"
     tags="$(echo $tags | sed -E 's![a-z/A-Z12]*\.java: (protected |private )?(static )?(final )?String [A-Z_]* = "([^\"]*)";!\4!g')"
-    comm="adb logcat $(echo "$tags " | sed "s/ /:$1 /g")*:S"
+    tags="$tags dalvikvm System.err AndroidRuntime "
```
menny: these are really good additions.

scripts/adb.sh
ArenaL5: I'd probably have tried your idea if I came up with it. But this was more or less what I was already doing by hand and didn't think of any other way.

Serendipity saves the day again.

scripts/adb.sh
ArenaL5: I used it on Linux.

I don't know how compatible is macOS with Linux, but if there are `bash`, `grep`, and `sed`, I don't see why not. I don't have a Mac device to test.

It won't work natively on Windows, but maybe with Cygwin and Linux emulation it could. Writing a separate script still seems like the safest bet.

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
ArenaL5: If you don't understand the problem, there was probably none to begin with. I think you added that line long ago.

I don't know why you needed to discard "expected events", I suppose it was just a heuristic to make the keyboard run quicker, and it went wrong.

app/src/main/java/com/anysoftkeyboard/keyboards/Keyboard.java
ArenaL5: Probably. I didn't really write that code.

I just needed to implement it at `Keyboard.java`, so I moved it from `app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java`, made it public and non-static. It seemed a better solution than copying it.

app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
ArenaL5: I'm not sure I understand your question.

`getMultiTapCode` naturally knew `mTapCount` when it was living at `PointerTracker.java`. It lost access to that variable when I moved it to `Keyboard.java`, so I'm passing it as an argument now.

I don't remember why did I make that method non-static. It was probably not strictly necessary, but it seemed more logical when put next to the other functions at `Keyboard.java`.

(If you're talking about the `if (multiTapCode < 32)` condition, it's because when it was a `char`, it could never be negative, so the line creating a String from that codepoint would never crash.)

scripts/adb.sh
ArenaL5: Lesson learned the hard way.

app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
menny: I think I did that since I assumed that if the event has not been delivered after 1500ms, it will never arrive....

menny:
@ArenaL5 can you close the relevant issues?

AnXh3L0:
I'll close them :)

Issue #1898, opened 2019-12-04
No backspace in first symbols keyboard


ashirviskas:
### Steps to reproduce
1. Click on "123!#()" button while in keyboard

### Expected behaviour
I see all the symbols and a backspace button

### Actual behaviour
I don't see the backspace button, just an empty space in its place.

**Android OS version:**
9.0
**Device manufacturer and model:**
Oneplus 6
**List of installed add-ons (like languages, or themes):**
Lithuanian and Russian keyboards. Nothing else.

![Screenshot_20191204-222921__01](https://user-images.githubusercontent.com/11985242/70178978-d9770800-16e5-11ea-8ac8-1a2b1acbf9c5.jpg)

ashirviskas:
Version 1.10.1064, version-code 6234

menny:
Thank you for reporting

ArenaL5:
I was hoping that no one would notice until I fixed it... that one was a typo I made. I accidentally disabled that key. Sorry.

I've made a fix (040c74d8daa3eee8d96880cb1ba37299b399ca22), and we'll merge it very soon, along with simple fixes for other bugs.

Issue #1798, opened 2019-07-31
[Gesture typing] reset word mode on cursor move/positioning


ildar:
### Steps to reproduce
1. Gesture type aword 
2. Tap in the middle of the typed word to move and position the cursor 
3. Preform anything, e.g. tap ‘a‘ to insert a letter
4. See insertion failed adding new letter after the word

### Actual behaviour
See the word underlined. It brings a lot of weird effects, especially in interference with neighbour words

menny:
@ArenaL5 can you take a look at this?

ArenaL5:
I'm testing on current master (commit cad0bdb35fe4eeea260bdff1f4a8cb67e3a05914). The only weird effect I see is that the inserted letter is pushed away:

1. The predicted word is committed.
2. A space is inserted. The cursor is put after that space.
3. The letter you tapped is inserted. 
4. The cursor moves like normal, and ends up after that letter, ending with (for example) "testing a".

It doesn't matter if I use the cursor keys or tap directly in middle of the word.

Of course it's still wrong behaviour, but it seems different from what @ildar is describing. (That's to be expected, as this report is from months ago.)

Could somebody verify from current master?

ildar:
I would be happy if anything was fixed. Then I'd test and report further
problems if remain
🤗
(As i already wrote I'm still on Gboard because of the annoying bugs that's
why I'm annoying )))

menny:
I see the behavior @ildar is describing on the latest _master_

1. Swipe a word (say, "what")
1. Click the middle of the word to move the cursor to after the "a"
1. Type another "a".

Expected to see "whaat", but I get “what a".

@ArenaL5 do you see a different behavior? @ildar is that the issue you see?

ildar:
Exactly. My version is 1.10.606 from
https://f-droid.org/app/com.menny.android.anysoftkeyboard

ArenaL5:
We're all seeing the same issue, then.

I'll look into it.

ildar:
I'm testing it and like it very much! Thanks a lot!

Issue #1797, opened 2019-07-31
[Gesture typing] ensure spaces on both ends


ildar:
When gesture is finished please insert a space before and after the typed word if neighbour words adhere

xdumaine:
Example: "have anice day" - no space was added between "a" and "nice"

ildar:
Today is a very nice day! I'm testing it and like it! Thanks a lot!

PR #1901, opened 2019-12-05
Small fixes

ArenaL5:
This pull request is smaller than my previous ones. but I'm already sending fixes at an ever slowing pace. I'd rather make these ones available (and close in the process my own blunder #1898) than making you wait upwards to one or two months.

These days I'll probably send very few patches, or focus in ones I find easy to fix like tweaking language pack layouts.

The commits are pretty self-explanatory, I only want to make a couple of comments:
- I changed a logcat tag and fixed a logcat message (that wasn't interpolating properly)
- #1798 (and thanks to your automated tests, #1797) are fixed in the most simple and literal way I could (changing `State.PERFORMED_GESTURE` to `State.IN_WORD` when moving over the word). I did some tests and they seemingly work. There's a lot to be improved, though.
- I also found a condition that may crash AnySoftKeyboard when gesture typing is activated Commit c3796c245412f25785ced7c74a08b47638e8c063 will catch that condition and safely output a warning to the logcat, although I don't know what can cause it.


app/src/main/res/xml/ext_kbd_bottom_row_simple.xml
```diff
@@ -7,7 +7,7 @@
     <Row android:keyWidth="10%p" android:keyboardMode="@integer/keyboard_mode_normal" android:rowEdgeFlags="bottom">
         <Key ask:isFunctional="true" android:codes="@integer/key_code_keyboard_mode_change" android:keyEdgeFlags="left"/>
         <Key ask:isFunctional="true" android:codes="@integer/key_code_mode_alphabet" ask:showInLayout="if_applicable"
-             ask:longPressCode="@integer/key_code_cancel" android:keyIcon="@drawable/sym_keyboard_cancel"
```
menny: we should add a checkstyle rule to disallow the use of this attribute.

menny: https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/1902

PR #1920, opened 2019-12-19
AnyKeyboard: Combining characters aren't EOW characters

### LINKED ISSUE #1919: Unicode Combining Characters Interpreted as End-of-Word Character
While I was trying to [make a Sinhala language pack](https://github.com/AnySoftKeyboard/LanguagePack/issues/277), I realized that [unicode combining characters](https://en.wikipedia.org/wiki/Combining_character) are interpreted as an end-of-word character (like a space character would be), resulting in the suggestions dictionary picking the word it thinks is closest at that point, as it would if you were to hit space.

I tried looking through the code, but I couldn't find the offending line. I know you've been working on code near this problem @ArenaL5 , have any ideas?

martholomew:
Fixes #1919 , thanks for pointing me in the right direction, @ArenaL5 !

Uses [getType](https://developer.android.com/reference/java/lang/Character.html#getType(int)): a return value of 6 is a [combining non-spacing character](https://developer.android.com/reference/java/lang/Character.html#NON_SPACING_MARK), a value of 8 is a [combining spacing character](https://developer.android.com/reference/java/lang/Character.html#COMBINING_SPACING_MARK).

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
+++ b/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
@@ -340,7 +340,7 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardColorizeNavBar {
     @Override
     protected boolean isAlphabet(int code) {
         if (super.isAlphabet(code)) return true;
-        // inner letters have more options: ' in English. " in Hebrew, and more.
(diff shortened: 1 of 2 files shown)


menny:
wonderful!

@martholomew, Can you add some unit-tests for this?

menny:
I think this should go to the release branch. Can you point your PR to `release-branch-1.10.-r3` rather than `master`?

martholomew:
Will do

Issue #1914, opened 2019-12-14
Word suggestions are always lower case


gartenriese2:
### Steps to reproduce
1. Use German dictionary. 
2. Type a word.
3. Press space.

### Expected behaviour
Suggestions for the next words appear and are spelled correctly.

### Actual behaviour
All suggestions appear in lower case.

**Android OS version:**
Android 10

**Device manufacturer and model:**
OnePlus 7 Pro

**List of installed add-ons (like languages, or themes):**
German dictionary

PFischbeck:
I found the source of this behavior. It happens because the typed words are already saved in lowercase for the next word suggestions:
https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/0c1df540ba84465c1e5be1de9fe3d1373624f6f6/app/src/main/java/com/anysoftkeyboard/dictionaries/Suggest.java#L224
The current word is lowercased before it is passed to the `mSuggestionProvider`, which then passes that lowercase word on to the `NextWordDictionary`.
https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/cad0bdb35fe4eeea260bdff1f4a8cb67e3a05914/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java#L459

AnXh3L0:
Thank you @PFischbeck for debugging this, maybe @ArenaL5 or @menny can implement the change?

ArenaL5:
Thank you for your help! It's really easy to fix when something tells you what to change.

I'll include this commit in my next pull request, hopefully this week (I'm trying to fix another issue and I've only done part of the work.

How come you didn't send a pull request yourself, @PFischbeck?

PFischbeck:
@ArenaL5 I currently don't have the project set up locally. Additionally, I wasn't sure whether just removing `.toLowerCase(mLocale)` would be okay, as this might have some side effects.

Currently (I think), if I type `This is`, then `this->is` is saved as a possible next word suggestion.
With the change, this changes to `This->is`, meaning that not only is the next word case-sensitive, but the current word too. In languages like German with many Capital letters, this will probably improve things (since the next words of a noun and a verb/adjective should be quite different, even if they are the same in lowercase). However, in most languages, this probably means the next word prediction after the first word of a sentence will be bad, or just non-existent.

In terms of design, it would probably be best to pass the words to the suggestions provider unchanged, and then at some later point one could think about handling this in the next word logic.

[...]

Issue #1662, opened 2019-03-14
Can't access apostrophe in address bar


n-o-o-b:
I can't access the apostrophe when typing in the address bar of my browser. Long-pressing any of the punctuation doesn't give me access to it and it's not in either of the symbols sections.

![Screenshot_2019-03-14-15-01-23](https://user-images.githubusercontent.com/19697175/54370946-74b7e100-4670-11e9-8dfd-7f26afc43e49.png)

Android 8.0, Firefox browser.

n-o-o-b:
Same goes for the semicolon.

Issue #1418, opened 2018-07-11
Replace extra ¿ with ⸘


gappleto97:
Currently it has two inverted question marks (¿), and an interrobang (‽), but no inverted interrobang (⸘).

It would be nice if the extra ¿ was replaced with ⸘.

gappleto97:
Please do not close this.

gappleto97:
Thank you!

Issue #535, opened 2015-11-15
Extension keyboard is right to left


il3li:
Hi,

It happens sometimes that an extension keyboard appear right-to-left in portrait mode (see an image below)
![screenshot_2015-11-15-14-49-13](https://cloud.githubusercontent.com/assets/15856890/11168895/8a8d64c2-8b71-11e5-8f54-b26c817aa5b8.png)

digi-owl:
Seems to happen when the swipe up is from the right half of the keyboard.

menny:
Yes, this is correct - when a popup keyboard is shown, there is a constraint that is checked: the first (highest priority) key in the keyboard should be just above the finger. Since the highest priority is the "1" key and there isn't enough screen space to draw the rest of the keyboard (keys 2->0), the keyboard view flips.

This makes sense in normal popups (long pressing L, for example), but I can see why it doesn't make sense in the extension keyboard.

I'll move this to the backlog.

ArenaL5:
I can't reproduce this bug in version 1.10.606 or the latest debug version, even when forcing right-to-left layouts.

I think this bug has been fixed by now.

PR #1921, opened 2019-12-19
Ability to reorder popup characters.

### LINKED ISSUE #1681: [Feature request] Add ė, į, ų letters.
These are the characters that hold me back from switching to Ask.

ArenaL5:
This pull request is mainly to close the issue discussed at the end of AnySoftKeyboard/LanguagePack#231. This will immediately fix AnySoftKeyboard/AnySoftKeyboard#1681 and, with time, several issues in both LanguagePack and AnySoftKeyboard repos.

It's... underwhelming, because @martholomew had the perfect interface for configuring this:

> These would all be _re-orderable checkboxes_ ([edit] not sure the exact name, but its the thing that looks like [this](https://www.youtube.com/watch?v=Hxc7l06xhv4)) in settings.
> I think that this would solve all sorts of problems people have.

But I realized I don't have the ability to implement that. The closest preference class in the Android libraries is MultiSelectListPreference, but that doesn't save the order of the checked elements. Writing event listeners for Android is already complex, and the drag-and-drop mechanism that @martholomew describes is so complex that several Maven libraries exist to implement it. And even then, we'd still have to add the checkboxes.

As I had already work half-done and I'd rather not delay this upwards of month, I implemented a very basic EditTextPreference, as seen in commit c7ba07817250636e541013d9b5987702bddd90b2. You can configure it from UI → Tweaks and more → Order of popup characters. I hope that this preference will be replaced for a more user-friendly one.

The instructions are very short but I think you devs will understand it pretty easily. The preference value is not case-sensitive, characters other than N, L, O, S have no effect, an empty string will disable popupCharacters altogether, and any duplicate character that would remain in the popup after reordering them all is discarded.

Besides all that, there's also fixes for some other issues, as you can see in the commit summaries (#1914, #1418, #1662) and a small addition to `README.md`.


ArenaL5:
Sorry, I always forget to use the automated tests.

The entirety of `testShortPressWhenNoPrimaryKeyAndPopupCharactersShouldShowPopupWindow` is throwing assertion errors because I broke something (I suppose with my first commit) and I hadn't noticed while testing it manually because it's not really a common occurence: a key with popupCharacters but no keycode.

I'll try to fix it soon.

ArenaL5:
I don't know if I made it clear, this works now. I fixed the bug I created with my last commit.

Is there anything left to fix before merging this?

menny:
@ArenaL5 would you mind trying to rebase this?

ArenaL5:
No problem!

ArenaL5:
This should do it. I've done all the automated tests and I tested it manually. `./scripts/ci/ci_check.sh` throws an error because it finds duplicate IDs I didn't edit. (for dictionaries and keyboards)

Everything else is okay.

nicoursi:
Hi! I was curious about this commit, so I compiled it to test it on my device. I was surprised to see how this commit breaks a symbols layout.  

As an example, try to use qwerty with symbols. You'll notice that none of those symbols designed for that layout will be available in the supposed order. I tried to change all the combinations with the letters `NLOS`, but there was no way that layout could be useful to me any more.
I don't know if I just didn't grasp its functioning or if this is really what this fix does.

If I may give you my opinion, I think `L`, or a new letter, should leave the active layout as it is. And then the user can play with the other letters if he wants to change it.

I notice, though, that it doesn't afffect long press popups like this one:

`<Key android:codes="101" android:popupKeyboard="@xml/popup_qwerty_e"/>`

So, if this PL is going to be merged, I hope there'll be a letter to preserve the current layout popup caracters order ;)

ArenaL5:
Sorry for the late reply.

`popupKeyboard` are custom layouts which can even have pre-recorded phrases, and my code can't analyze them, so they stay as they are.

My patch is only supposed to reorder symbols in `popupCharacters`, and no, I didn't think of adding an option to leave them as they are. Did you intend to have symbols at both sides of the pop-up or something?

About it not working, it's possible, I haven't tested it on a long time. I'll rebase this branch and try to fix it. Thank you! I'll let you know when I know something.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
```diff
@@ -674,13 +674,16 @@ protected boolean setupKeyAfterCreation(AnyKey key) {
             return true;
         }
 
-        // filling popup res for external keyboards
+        // if only popupCharacters are specified, call super nevertheless --
+        // we might want to add more characters.
+        /*
```
menny: I see that this logic has moved to a different class. Please remove this from here completely, we don't need it.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/ExternalAnyKeyboard.java
```diff
@@ -414,122 +431,174 @@ public boolean isInnerWordLetter(int keyValue) {
...
-        // ΑΒΞΔΕΦΓΘΙΪΚΛΜΝΟΠΨΡΣΤΥΫΩΧΗΖ VIM DIGRAPHS
-        // αβψδεφγηιξκλμνοπ;ρστθωςχυζ Greek layout
-        // ΑΒΨΔΕΦΓΗΙΞΚΛΜΝΟΠ;ΡΣΤΘΩΣΧΥΖ GREEK LAYOUT
-        // αβχδεφγηι κλμνοπθρστυ ωχψζ Magicplot
-        // ΑΒΧΔΕΦΓΗΙ ΚΛΜΝΟΠΘΡΣΤΥ ΩΧΨΖ MAGICPLOT
+        /* We're assigning greek letters to latin letters
+         * by mapping the standard greek layout on a PC to a
+         * QWERTY layout.
+         * Comparison of several mappings:
+         * ABCDEFGHIJKLMNOPQRSTUVWXYZ QWERTY KEYBOARD
```
menny: beautiful!

ime/app/src/main/res/values-es/strings.xml
```diff
@@ -349,6 +349,9 @@
     <string name="bottom_generic_row_group">Fila inferior genérica común</string>
     <string name="bottom_generic_row_dialog_title">Seleccionar el tipo de fila</string>
     <string name="bottom_generic_row_summary">Fila inferior seleccionada: %s</string>
+
```
menny: Translating should be done in https://crowdin.net

menny (changes requested):
Great! I love it.
A few comments and also, please, add some unit-tests.
And add an entry in the change-log [here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/java/com/anysoftkeyboard/ui/tutorials/VersionChangeLogs.java#L22) and [here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/play/release-notes/en-US/alpha.txt).

nicoursi:
> **-------------------------edit----------------------------------**
> I made some more tests, and I think that it's not working at all. It doesn't matter how I change those letters order, the popups always appear at the same order. I don't know if this is device specific or if something got broken by this: `Merge branch 'master' into patches`
> **-------------------------edit----------------------------------**
> 
> Hi! I was curious about this commit, so I compiled it to test it on my device. I was surprised to see how this commit breaks a symbols layout.
> 
> As an example, try to use qwerty with symbols. You'll notice that none of those symbols designed for that layout will be available in the supposed order. I tried to change all the combinations with the letters `NLOS`, but there was no way that layout could be useful to me any more.
> I don't know if I just didn't grasp its functioning or if this is really what this fix does.
> 
> If I may give you my opinion, I think `L`, or a new letter, should leave the active layout as it is. And then the user can play with the other letters if he wants to change it.
> 
> I notice, though, that it doesn't afffect long press popups like this one:
> 
> `<Key android:codes="101" android:popupKeyboard="@xml/popup_qwerty_e"/>`
> 
> So, if this PL is going to be merged, I hope there'll be a letter to preserve the current layout popup caracters order ;)

Thanks for answering back. I tested it again after rebasing.


About not working. I found out it does indeed  work but changes in the string  do not happen straight away. They take effect after the app is cleared from recents. This can be very confusing.

If someone uses some of the layouts with symbols (eg. QWERTY with symbols), after an ASK update (including this PR) he won't have easy access to those symbols on long press. And changing the settings won't work unless he clears everything in recents.

With NSLO I get most of the keys right except:

E: supposed to give 3 but gives ₃
T: supposed to give 5 but gives ț
U: supposed to give 7 but gives ű
C: supposed to give  - but gives ĉ

I hope you find this discussion useful, otherwise just let me know and I'll leave you alone with your tests 🙂

ArenaL5:
@menny. Thanks for the compliment but I didn't do anything yet. I just rebased again to clean up the history. :laughing: I'll try to fix those issues soon.

@nicoursi: It was useful in fact. I did notice that the keyboard retained its old settings for a while, but I didn't know what to do to load the new ones.

Options programmed for other people do take effect straight away, so there's probably a line of code to flush the cached keyboard. I'll investigate.

menny:
Yes, there is a way to flush the cache. Look for other places where we react to preference changes (in the AnySoftKeyboard* classes)

ArenaL5:
I just rebased again, and I took the chance to squash some commits and to fix some weird spots in the history (comments that don't belong, etc.), so the commits @menny referenced in the review are now tagged as outdated. I removed the translation (it was its own commit) and the outdated code.

The thing now updates its configuration instantly when changing preferences (commit 7e8caf0d0992325d0b2a30cba06c403988a2532a; @nicoursi, please test).

I also updated the changelog in commit 087683e3089996220ecd4fe73fa739c0b676c421. @menny, is this okay?

(I haven't done any test units yet, I'll do them later.)

nicoursi:
@ArenaL5 
Tested :)

Flashing works all right now. 

There is still the problem about layouts with symbols. The default string breaks them. After a few trials and errors, I found that the string `NSL` fixes it. `NSLO` breaks a few keys (see my comment above).

I think this is confusing and might raise a lot of bug complaints. Finding the right string is not straight  forward. I also think `NSLO' should not break symbols layouts. I also think it should be advertised the fact that you can also remove some letters from the string.

Here's my two cents ;)

ArenaL5:
Sorry I completely forgot about your issue. :sweat: Yeah, I just tested it now.

open_popup_from_the_middle

The rightmost keys fill the popup from right to left, every other from left to right. The number is correctly on the left side of the popup, but the popup is centered so the finger automatically selects a random letter in the middle.

I can mitigate that bug, but the only ways to completely fix it I can think of are:
- reordering the letters so that the first symbol is always under the finger (but then the rest rearrange themselves in a confusing order), or
- constraining the width of the popup to six characters only (if the key is perfectly centered like V, shift the popup some pixels to the left/right so it fits on the screen).

I favour the second one. @menny, would that cause any problem?

nicoursi:
What about splitting them in two rows? Would it be feasible? 

If I am not wrong it already splits automatically at 10 letters, or even 9. So the 6th letter is going to be the 1st on the top row. And from 1 to 5 are going to be on the bottom row. Finding a way to make it split after 5, would be a solution I guess.

I can make some tests tomorrow to define what's the minimum number of letters to make it automatically split.

ArenaL5:
Currently it starts filling the second row when it reaches 8:

1 to 8 symbols: 1 row.
9 to 16 symbols: 2 rows.
17 to 24 symbols: 3 rows.
25 or more: (not defined, probably it will break)

(Unless you lay out you own popupKeyboard like you did [here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/1921#issuecomment-602790073), of course. Those are exempt.)

Your solution would be more comfortable than the current one. Maybe forcing it to be two rows when it reaches just 4 symbols?

1 to 3 symbols: 1 row
4 to 12 symbols: 2 rows
13 to 18 symbols: 3 rows

nicoursi:
I think it could work,  but we should take into consideration that (if the 4th is going to be the 1st top row)  position 1 and 5 are going to be high priority popup caracters. I think that selecting the 1st  top is faster than pressing the second bottom.  I don't know if this is for everyone or just for my typing habit. 

If this is for everyone,  the list of popup caracters should be rearranged according to that.

ArenaL5:
(This is for future reference. My previous post contained an innacuracy: if there are more than 24 symbols, it still outputs three rows. Maybe they overflow out of the screen -- I haven't tested it)

https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/522dab6089ce2642ee43d39df92f2c5cdd3ba2c0/ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyPopupKeyboard.java#L145-L166

> I think that selecting the 1st top is faster than pressing the second bottom. I don't know if this is for everyone or just for my typing habit.

It seems like your typing habit to me. Keys are less wide than high, so the finger should move less to the right than upwards; and we're already trained to stay in the lower part of the screen. Moving upwards while holding the phone from the bottom is more uncomfortable for me.

I can adapt either way, but I'm not sure what the majority will prefer. If we adopt your proposal, remember the width of the popup is not fixed (symbols are split evenly among all rows) so to make the code fully general, I think we should fill up the popup by columns:

```
2468
13579
```

But this means that all the symbols will be rearranged when we hit three rows:
```
369
258
147
```

(As you can see, I'm very opposed to just swapping two symbols and leaving the rest in the same order.)

nicoursi:
The way it is faster for me, is that I can long press and then swipe up to pick the fist on the top row. To select the second bottom I have to long press and then carefully select the second one. That's surely a habit ;)

ArenaL5:
That's a _very_ good point. I tried sliding upwards now and it's not as terrible as I expected. :laughing:

Okay, you've convinced me. When I find the time I'll push a commit trying your proposal. If the ASK team disagrees, I'll change it back.

nicoursi:
Yes swiping is nice :) My dream is to to implement swipe up and swipe down on every key. Swipe up (or down according to the implementation) would uppercase the character and swiping down (or up) would select the first popup character. I'm not skilled enough to do that yet, but I'm starting to look into it.

ArenaL5:
I'm sorry to say I don't have any advice for your dream, but if I stumble upon a good starting point for you, I'll tell you. :slightly_smiling_face: 

I just pushed a commit rearranging the popups as we discussed it. The first character should always be on the key now, and they should be laid out in columns. Would you test it?

(Also, please don't mind the force-pushes, I kept sending stupid format mistakes and useless code. Kind of embarrassing when the solution's so simple :sweat_smile:)

nicoursi:
I tested it. It's better now. I still see a problem though :(

This feature is certainly useful for those layouts that have not been specifically designed for a language. So, it's useful and necessary for generic layout.

But, if a layout has been already specifically designed for a certain use-case or language, then this feature might break the order symbols and letters are supposed to appear in terms of accessibility.

I'll give you a specific case: `Italian Qwerty with symbols`. It's been designed to have some symbols first, while keeping easy accessibility to those vowels with diacritics present in the Italian language: `à, è, é, ì, ò, ù`. Just these ones, no more!!

Let's have a look at the key `e`:
In the above mentioned layout you had: `3` and `€` as first choices for the bottom row, and `è` and `é `as first choices for the top row and then some more symbols. So it mixes letters and symbols

With this new feature (NSL) I get this:


Beside the fact there seems to be  a duplicate, it splits letters and symbols in two groups. So there's no way to have `è` and `è` close together with `€` and `3`. I wouldn't really be happy to give this up.

The easiest solution to this I can see, would be to have a checkbox to disable this feature all together for people who use a specific, and not generic, layout. I would still use the row spitting improvements for every layouts, because It fixes a common problem.

This is just my opinion, of course! :)

nicoursi:
The point is Italians (in the specific case) don't need `êëęē` at all. I'll see what happens if I delete them from the italian with symbols layout,  if they'll appear anyway or not. `êëęē` have no priorities at all for italians and maybe for  other languages as well, so I'm afraid we will lose some customizations we could arrange before with custom layouts. 

I'll do some tests and I'll report.

ArenaL5:
They should appear only if you set `O` in your settings. _Should._

Please do.

nicoursi:
> They should appear only if you set `O` in your settings. _Should._

I'll try. If it works all right I will edit the Italian Qwerty with symbols layout.

nicoursi:
Ok, I tested it again :) 
It indeed works well, but in order to work as expected and without discomfort for anybody I think all the layouts should be updated.

Every layout with symbols should have removed all the diacritics not belonging to that locale. And if somebody needs those ones, can use `O`.

Every simple layout should be also updated, making sure there are all the diacritics used by that specific language. Otherwise the string `L` as for `Locale` would not make much sense.

In my specific case, I added `è`, `é`, `ò`, `à`, `ù` to the normal `Italian qwerty layout`, and deleted every diacritics other than those `italian` specific ones from the `qwerty with symbols`.

Right now, English with Symbols, Spanish with Symbols, French with symbols, and probably many otherr have diacritics which don't belong to those languages, and might create discomfort to the users because they are treated as local diacritics (L) but they are not.

This should be investigated I think.
One thing I also ask myself is: is `NLSO` really what most people need? Who needs all those symbols and diacritics all together? :) Wouldn't mabybe, just `NLS`, be a good default string? If the standard locale layout is well designed, there will be no need for 'O' for most people, I think.

Another thing I notice is that for the key `h` with the `O` string I only get `ĥ` and a longer n, no idea what that is called. But `ẖ`, `ḫ` and  `ḥ` are missing. Any reasons for that?

nicoursi:
Also `hintLabels` will not be loyal anymore. In many cases, expecially with  the default string, or also on the case of `a` from the example above, what you see on the hint label won't be the first choice on longpress.

ArenaL5:
> Ok, I tested it again :)
> It indeed works well, but in order to work as expected and without discomfort for anybody I think all the layouts should be updated.
>
> [...]
>
> Every simple layout should be also updated, making sure there are all the diacritics used by that specific language. Otherwise the string `L` as for `Locale` would not make much sense.

That's the idea hahah. Once I finally add the tests and this gets merged (and maybe this is published in F-Droid and Google Play), we should remove all clutter from every layout.

(By the way, I need to investigate why that test is failing with the last commit. The test fails every time and only with the last commit, but it's reporting a symbol I don't see when testing it by hand. I must be looking at the wrong layout or something.)

> One thing I also ask myself is: is NLSO really what most people need? Who needs all those symbols and diacritics all together? :) Wouldn't mabybe, just NLS, be a good default string? If the standard locale layout is well designed, there will be no need for 'O' for most people, I think.

You're probably right. The way the application worked was pretty similar to hard-coded `NLS`, after all.

> Another thing I notice is that for the key `h` with the `O` string I only get `ĥ` and a longer n, no idea what that is called. But `ẖ`, `ḫ` and `ḥ` are missing. Any reasons for that?

It's eta. A Greek letter.

https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/d0253bb22f3b8152c3226366be4f3cf7fa790fc7/ime/app/src/main/java/com/anysoftkeyboard/keyboards/ExternalAnyKeyboard.java#L469-L471

They're not part of the general defaults and they weren't before I started changing the file. They probably were defined in some other local layout...

A Wikipedia search doesn't give me any result for `ẖ`. I'm not sure it's in common use in any language, besides specialist texts... But `ḥ` is (in Asturian and others) and `ḫ` is too (in Arabic transliteration), so I suppose I'll add them.

> ----- this is another update -----
> The case of mixing symbols and L is still present.
> This is an axample:
> 
> `NSL`:
> long press on `a`: `@` first choice and `à` as second choice as expected.
> long press on `e`: `3` and `€` as first choices, then `³` and `₃` and as last choices `è` and 'è`.
> 
> `NLS`:
> long press on `e`: `3`,` è` as first choices as expected
> long press on `a`: `à` first choice and `@` as second choice. Not as expected by the designer.
> 
> As it's a `symbols` layout, symbols and numbers have priority and local diacritics should be second choice.
> But there is no way to accomplish that in this case as `@` should be first choice before the diacritic. But doing that it breaks `e`. I understand it would be a lot of work, so I guess I will have to remove `³` and `₃` completely from the Italian symbols layout so that I' don't break other priorities. `@` on long press is a must for me, as it is `è` and `é` after `3`. I am too used to it to give it up :)

https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/df12e0f5e7d19fe894e758b873735102c2de71d2/addons/languages/italian/pack/src/main/res/xml/italian_qwerty_with_symbols.xml#L11
https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/df12e0f5e7d19fe894e758b873735102c2de71d2/addons/languages/italian/pack/src/main/res/xml/italian_qwerty_with_symbols.xml#L22

Yeah, I see the problem. It sounds incredibly lazy coming from the creator of this PR, but removing the threes would be the easiest fix. :joy: (They're defined as popups in the number keyboard, if that's any consolation.)

There may be an easy solution, but I'm not sure which. Adding classifications for symbols is trivial, it's just adding some constants and tweaking this filter:

https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/d0253bb22f3b8152c3226366be4f3cf7fa790fc7/ime/app/src/main/java/com/anysoftkeyboard/keyboards/ExternalAnyKeyboard.java#L533-L546

and I'm willing to add a couple of groups more (like a G for "Greek letters"), but I'd rather avoid adding a myriad groups to avoid taxing the users. If you can think of any simple criterion to separate the at-sign from the small numbers (that you think you'll be useful for someone else), please tell me.

nicoursi:
> If you can think of any simple criterion to separate the at-sign from the small numbers (that you think you'll be useful for someone else), please tell me.

What about creating a `currencies+@` group? This would make sense for people like me that don't use the symbols keyboard at all but only a layout with symbols. This way they can have easy access to their currency symbol. I would see no harm adding the @ to this group, but I'm not sure.

This group would surely save the ³ and ₃ from Italian QWERTY with symbols 😆

nicoursi:
<A Wikipedia search doesn't give me any result for ẖ. I'm not sure it's in common use in any language, besides specialist texts... 

```
I found this:
Sometimes used for Arabic خ ẖāʼ, Hebrew Heth (letter), Egyptian 𓄡.

There is no precomposed upper case equivalent of ẖ so it uses a combining macron below instead: H̱. 
```
https://en.m.wikipedia.org/wiki/Macron_below
https://www.lexilogos.com/keyboard/arabic_latin.htm

ArenaL5:
Let me say that now we're at this point, we'll need a user manual for sure. With a section for advanced users, and a section for "I don't care, just give me the default" users.

By the way, I'd like to ask @martholomew for his input, too. We discussed this in the past (AnySoftKeyboard/LanguagePack#231) and he'd probably would be interested.

> What about creating a `currencies+@` group? This would make sense for people like me that don't use the symbols keyboard at all but only a layout with symbols. This way they can have easy access 
to their currency symbol. I would see no harm adding the @ to this group, but I'm not sure.

I'm not convinced... most people will only use a couple of symbols for currencies, maybe three. Also adding @ seems arbitrary.

> Another idea that comes to my mind is adding a single character group. Let's say, if somebody wants easy access to a specific character, he could add `@` in the string, to give a different priority to the `@` symbol. This could be any character, @ was just an example. Any character other then the ones already used for the string (NSLO) would create a single character group. `@` for @, `c` for c, `~` for ~ and so on. this would add a big deal of customization. I think this should be easy to implement, and potentially one could use multiple single character groups

I actually like this idea. So much. It's a simple way to address very specific needs without creating a whole new layout.

Just remember this string applies to every key in every layout.

> I was thinking also that whatever showing on the `hint label` should always appear first on long press. Otherwise, `hint labels` will not make sense any more. And if you introduce `single character groups`, you will have to decide what's going to have precedence: `hint label` or `single character group`.

Sorry, I forgot to address this in my last post.

I think hint labels are supposed to be automatic, but I don't remember having updated that mechanism so I'll change it. Hard-coded hint labels are still allowed and that doesn't make sense with customizable order.

Hint labels should be rewritten every time you change your layout or your preferences. If @ is the most important symbol for your key, `@` should be the first in the hint label.

> Another thing I'm thinking of is that every layout might need a different string according to how it's been designed. And to make it even better, the layout designer could put in the xml the preferred string for that layout. And if none, the default would apply. And the user can always change it. I know this is more complicated to implement, but I'm brainstorming here :smiley: This could at least be a future enhancement.

This doesn't... look sane. :laughing: It _could_ be implemented, but what kind of special layouts are you planning on using this for?

Symbols in a category aren't reordered: if a key has popupCharacters `áä` and another one has `äá`, both will remain in that order unless you use a one-character group. Unless you plan on mixing symbols and letters very thoroughly, the one-character groups and the current settings should be enough for pretty much any configuration.

If that's not enough, I'm warming up to your idea of allowing you to disable the reordering. Or even better: to add an extra fake category: `X`, meaning "Import the rest of the symbols in the layout, without reordering them". The current code deletes repeated characters anyways, so you can still override the order for some symbols by setting your preference to `NX` or `@X` or `XO`

ArenaL5:
Again, there are so many things to discuss that I forgot about your later post:

> > A Wikipedia search doesn't give me any result for ẖ. I'm not sure it's in common use in any language, besides specialist texts...
> 
> This diacritics for `h` are present in all `with symbols` layouts and I wasn't allowed to remove them from `English with symbols`. That's why I asked you, so that they can be eventually removed from `English with symbols` layout.

You weren't allowed? I suppose the maintainers might have been afraid of you breaking something... but when this is tested and merged, it will only make sense to remove them.

> I found this:
> 
> ```
> Sometimes used for Arabic خ ẖāʼ, Hebrew Heth (letter), Egyptian 𓄡.
> 
> There is no precomposed upper case equivalent of ẖ so it uses a combining macron below instead: H̱. 
> ```
> 
> https://en.m.wikipedia.org/wiki/Macron_below
> 
> https://www.lexilogos.com/keyboard/arabic_latin.htm

Then it's better to remove this one. The popupCharacters mechanism is only compatible with precomposed characters.

To be exact, it will split the sting in codepoints. `ẖ` is one codepoint, while `H̱` is two. For that reason, it can't uppercase the letter no matter how you press the Shift key, and if you try to add `H̱` it will split it into `H` and `̱`.

I expect less people missing that character than people asking "why won't uppercase work for H?".

nicoursi:
> I think hint labels are supposed to be automatic, but I don't remember having updated that mechanism so I'll change it. Hard-coded hint labels are still allowed and that doesn't make sense with customizable order.

Currently, hint labels are indeed automatic but they make the keyboard less clear because all of the available pop up characters will be shown as hint labels, and this makes it really difficult to distinguish them, as the font size is really little.

I usually hardcode hint labels so that they will never be more than 2. Usually I only use one hint label per key, sometimes two.

nicoursi:
> I'm warming up to your idea of allowing you to disable the reordering. Or even better: to add an extra fake category: X, meaning "Import the rest of the symbols in the layout, without reordering them". The current code deletes repeated characters anyways, so you can still override the order for some symbols by setting your preference to NX or @X or XO

I'm not really sure I understand this. You disable reordering only for some symbols?

The best way to really test this is to design a couple layouts and if you don't get any headaches you are good to go. I'm glad I designed two layouts, and I was really frustrated because I had to delete a couple of characters from it, because they would always be put at the wrong place. It might get fixed by single character group, but meanwhile, I had a look into the code and disabled reordering with `X `at the beginning of the string. This doesn't give me any headaches at the moment. Feel free to use it, modify or not use it at all if you wish. Have a look at it here: [e4f3106 ](https://github.com/AnySoftKeyboard/AnySoftKeyboard/commit/e4f31060c17223bb3035be5894fa2df71260821a) 
I think this ability to disable it should be present though. If the designer has something specific in mind, the user can put an X in the string, otherwise he can use different strings. 

>This doesn't... look sane. laughing It could be implemented, but what kind of special layouts are you planning on using this for?
Symbols in a category aren't reordered: if a key has popupCharacters áä and another one has äá, both will remain in that order unless you use a one-character group. Unless you plan on mixing symbols and letters very thoroughly, the one-character groups and the current settings should be enough for pretty much any configuration.

This happens more often than you think. I encountered this problem in different cases. Sometimes one string would break the same layout (fixing one key but breaking another) and other times it would fix one layout but break another. Sometimes you want local diacritics to appear first on a symbol and be followed by (e.g.) a currency symbol. On another key you might want the `<` sign first but you also have the `µ` math symbols (it's not always used as a letter) that you want with less priority. At the same time you might have similar cases on another layout but in reversed order.

Advising in the XML if a specific layout should be best used with X (or any other string) is a perfect way to handle the whole thing IMO. It is more difficult to implement, of course,  as you would need to use a different string for each layout, but at the end you get a fully customizable feature.

ArenaL5:
Sorry for the delay. As always, some days I can't find the time to check this, but I really should at least say something before I disappear for so many days. (I'm really sorry this is taking so long, but it'll be worth it, I'm sure.)

Your commit looks good, disabling reordering is pretty simple. I'm still not sure if many people would use a setting to adopt the recommended designer's order, but I suppose I'll add the option as its own commit and see what happens. I'll still add the other categories first.

Disabling reorder for only some symbols would look like this (it's much easier to show you than to tell it in words):

- Suppose a key has `3eéε,` as popupCharacters.
- Your `X`, like you implemented it, would create a key with `3eéε,` (The same order.)
- A setting like `XG` or `GX` would only move the greek letter(s), but all the other symbols would remain in place: `3eé,ε` or `ε3eé,`

(Or as the script does: each uppercase letter stands for a subset of the characters, and after the whole preference is processed, any duplicate character is deleted. `G` stands for `ε` and `X` stands for `3eéε,`, so `GX` would be `ε3eéε,` which would become `ε3eé,` after deleting duplicates.)

ArenaL5:
I couldn't write new working code yet, but I found something worthy of an update. Maybe you know how to deal with this painlessly:

While I was thinking about adding a Greek category, I found the Google interface for [`Character.UnicodeScript`](https://developer.android.com/reference/java/lang/Character.UnicodeScript), and I realized I was only testing for Greek letters in Latin keyboards.

That interface comes with a method `Character.UnicodeScript.of()` that basically retrieves the name of the writing system where a character belongs. (If you feed it an alpha, it will output a constant meaning GREEK; a G, LATIN; a period, COMMON.) I figured the easiest method to generalize the Greek letters category was to compare the results of `Character.UnicodeScript.of(primaryCode)` and `Character.UnicodeScript.of(popupCharacter)` to detect "foreign" characters (like a Greek letter in a Latin key, or a Latin letter in a Russian key) and separate that onto their own category: F (for "foreign").

The bizarre thing is that the code compiiles correctly (you just need to `import java.lang.Character.UnicodeScript`) but when I try to use ASK on my phone, it crashes with this message in the logcat: `Didn't find class "java.lang.Character$UnicodeScript" on path: DexPathList`. The target SDK is okay, the documentation is there, but the library just doesn't exist in my phone. This [has happened to other people too](https://stackoverflow.com/questions/41557329/java-lang-classnotfoundexception-didnt-find-class-java-lang-characterunicode).

I can't simply copy [content from Java SE](https://hg.openjdk.java.net/jdk/jdk14/file/4f2af72cb65a/src/java.base/share/classes/java/lang/Character.java) as it is licensed under GPLv2, while this is under Apache. Is there any way to use this library without licensing problems or must we parse [the Unicode raw data](https://www.unicode.org/Public/14.0.0/ucd/Scripts-14.0.0d1.txt) ourselves?

menny:
this is really powerful addition. But, unfortunately, it was added in API level 24: https://developer.android.com/reference/java/lang/Character.UnicodeScript
I don't think it is wise to use it right now in ASK

ArenaL5:
Alright then. It's not really a priority, and there probably is a better method out there.

I just pushed tests for this code. There are some tasks pending.
- I didn't define any tag for recommended layout yet (as @nicoursi recommended [above](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/1921#issuecomment-610588485)) but the `X` setting is implemented, which is more or less synonymous (it loads characters without reordering them, following the designer's recommended order).
The code could be improved. Right now, you can add one-character groups before the `X` (so `3X` would transform `e3é` into `3eé` but not after it (`X3` transforms `e3è` into itself).
- I didn't include any modification to the key labels yet. Automatic labels should reorder with the setting, but hard-coded key labels are still allowed; maybe we could let 1-character labels for bundled keyboards and forbid manual labels on language packs? Or reorder characters in hardcoded labels according to user's settings?
- The setting description should be updated. I'll later try to add a brief manual in the settings screen (and doesn't require the user to open a web browser).

By the way, I did a force-push changing some commits, and made up a keyboard parameter on commit 059532cc1193504b966da51bb361c15fa213f913 so we wouldn't get a repeat of #535 (which my previous code caused). Any feedback about it?

ArenaL5:
Finally, I think this is all done. I'm very sorry this took so long, I lost count of how many times I had to rewrite code and try and guess why the tests failed.

I had to rebase everything, amend a couple of commits, fix a bug I created (see 36f246d46eb6c5de7d9618d2d8aac2fcc6ebfdbe) and, most importantly, change one of your tests in commit 4e0201157d33e34ae9a05bfabfde55686c3405d4.

Also, you may want to double-check the changelog (commit 8e027a312ed75d7dc39948a94e7dae90d82f961f).

I can't remember anything else noteworthy. Please verify.

ArenaL5:
Okay, I've been using this branch for a couple days and I see no hint of memory leakage or any breaking. I think it's good to go.

@menny: **Merging this branch __will__ change how popupCharacters are laid out, as discussed [earlier](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/1921#issuecomment-605902796), but you have the final word on this.**

Currently, for keys on the left half on the keyboard, symbols in popupCharacters (after reordering) will be laid out in columns, bottom to top, then left to right.

```
3 or less characters: 1 2 3

4-12 characters:      2  4  6  8 10 12
                      1  3  5  7  9 11

13-18 characters:     3  6  9 12 15 18
                      2  5  8 11 14 17
                      1  4  7 10 13 16
```

For keys on the right hand, the rules are the same but the result is mirrored. **If you feel another order is better or want to add a user preference just in case, please say so.** If you find no problems, merge away!

Merry Christmas. :christmas_tree:

ime/app/src/main/res/values-en/strings.xml
```diff
@@ -485,6 +483,8 @@
     <string name="bottom_generic_row_dialog_title">Select row type</string>
     <string name="bottom_generic_row_summary">Selected bottom row: %s</string>
 
+    <string name="popup_characters_order_title">Popup characters order</string>
```
menny, line 486: you can remove these changes in the `en` locale, the github bot will fill those correctly

ime/app/src/main/res/values/settings_keys_dont_translate.xml
```diff
@@ -374,6 +374,8 @@
 
     <string name="settings_key_allow_layouts_to_provide_generic_rows">settings_key_allow_layouts_to_provide_generic_rows</string>
     <string name="settings_key_apply_remote_app_colors">settings_key_apply_remote_app_colors</string>
+    <string name="settings_key_popup_characters_order">settings_key_popup_characters_order</string>
+    <string name="settings_key_popup_characters_order_default">NLOS</string>
```
menny, line 378: the default value should be in `settings_defaults_dont_translate.xml`

ime/app/src/main/res/values/strings.xml
```diff
@@ -485,6 +485,16 @@
     <string name="bottom_generic_row_dialog_title">Select row type</string>
     <string name="bottom_generic_row_summary">Selected bottom row: %s</string>
 
+    <string name="popup_characters_order_title">Characters shown upon long press</string>
+    <string name="popup_characters_order_summary">Order is important. Uppercase letters represent different characters for every layout:\n
```
menny, line 489: How does this long explanation look in the UI? Can you share a screenshot?

menny (changes requested):
Another question: how does the `reverse` value affect the currency XMLs in the app? And did you do a quick survey of if/how this will affect XMLs in various language-packs?

menny:
Overall, this is a great change!

ArenaL5:
Thank you!

Sorry I've been a little disappeared lately. I'll push new commits to fulfill your requests. It should take a few days to close them all.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/ExternalAnyKeyboard.java
```diff
@@ -34,6 +34,7 @@
 import com.anysoftkeyboard.utils.LocaleTools;
```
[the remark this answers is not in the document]
ArenaL5: That's odd... that usually happens when I copy a file from a thumb drive or my phone, but I don't remember doing any of that...

Fixed in 3e717937c167800fdf042dc2f66ce7c4dc70e6f2. By the way, when we're done with this, I'll rebase (and squash these commits so it doesn't look like a giant mess).

ime/app/src/main/res/values/settings_keys_dont_translate.xml
ArenaL5: It is there. The default value is `NLOS`, line 378.

Is there any other `settings_defaults_dont_translate.xml`?

ime/app/src/main/res/values/strings.xml
ArenaL5: Sure, here. Please excuse the Spanglish.

Screenshot_20210107-183045_AnySoftKeyboard_Dev

I don't remember the exact details, but I had to tweak a template for an info section in XML. It must be at 380af81088c8494ef8917fda619f012027a6eaca (“Add better instructions for the characters reordering”).

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
```diff
@@ -675,13 +675,8 @@ protected boolean setupKeyAfterCreation(AnyKey key) {
...
 
-        // filling popup res for external keyboards
-        if (key.popupCharacters != null) {
-            if (key.popupCharacters.length() > 0) {
-                key.popupResId = com.menny.android.anysoftkeyboard.R.xml.popup_one_row;
-            }
-            return true;
-        }
+        // if only popupCharacters are specified, call super nevertheless --
+        // we might want to add more characters.
```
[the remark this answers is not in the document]
ArenaL5: If I hadn't double-checked, I've sworn I was restoring a comment from a previous version. But I can't find it in `master` history. Wow, this is badly worded.

I think I meant: “if popupCharacters are defined but popupResId isn't, call super anyways -- we might want to mix them with the default characters in ExternalAnyKeyboard.java”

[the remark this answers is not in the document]
ArenaL5: Comment rephrased in b728ef768a61592485ad914b42f9f8f2556cb09d.

Again, I _will_ rebase this and squash microcommits like these when everything's done.

ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardRxPrefs.java
```diff
@@ -147,6 +147,8 @@ protected void onSharedPreferenceChange(String key) {
                 || key.startsWith(KeyboardExtensionFactory.BOTTOM_ROW_PREF_ID_PREFIX)
                 || key.startsWith(KeyboardExtensionFactory.TOP_ROW_PREF_ID_PREFIX)) {
             onAddOnsCriticalChange();
+        } else if (key.equals("settings_key_popup_characters_order")) {
+            onAddOnsCriticalChange();
```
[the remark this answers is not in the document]
ArenaL5: Is this enough? ad47846d574ff47949d10d4fc30cdda8e805ab97

ArenaL5:
Answering your first question (which I didn't see at the moment):

> Another question: how does the `reverse` value affect the currency XMLs in the app? And did you do a quick survey of if/how this will affect XMLs in various language-packs?

I had only tried it with the Spanish layout, which only defines the main layout and popupCharacters (it doesn't include XML files for static popups). 

I'm testing now other keyboards:
- the Brazilian layout with no Ç or ' key: it reverses the popup for U, but this is consistent with the popups created from popupCharacters and this code)
- emoji keyboards: it reverses the popup for some emojis, but this is confusing because the race selection switches directions for ~40% of the emojis.

I will push an extra commit to fix the emoji problem and any other layout my code breaks.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyPopupKeyboard.java
```diff
@@ -157,12 +147,10 @@ private static int getPopupLayout(CharSequence popupCharacters) {
...
-        if (count <= 8) return 1;
-        if (count <= 16) {
-            return 2;
-        } else {
-            return 3;
-        }
+
+        if (count <= 3) return 1;
+        if (count <= 12) return 2;
+        return 3;
```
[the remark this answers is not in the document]
ArenaL5: `testCodesParsing()` at `ExternalAnyKeyboardTest.java` tests it indirectly.

I could add a test-unit to test this directly, but how do I test a private method?

ArenaL5:
I made more tests; apparently these commits won't break anything. The currency key in the main symbols keyboard is a normal key with popupCharacters on the left hand of the keyboard, so it doesn't even get mirrored. The main symbol is the € sign for me.

The keyboards using predefined popups (most if not all of them could be changed to popupCharacters) appear correctly. Popups opened on the right half are mirrored exactly as they're supposed to do, no matter if they're defined as XML or built up from popupCharacters:

old_rendering
mirrored

Unless you remember some other popup that should _not_ be mirrored (emojis were already taken care of), this looks good. Maybe the 16-keys keyboards, but the popups in that were already faulty (too uneven and often too wide). I propose myself to convert them to popupCharacters in a different PR; if you have a better idea, please tell me so.

I think I addressed everything now. If you notice something wrong, please tell me; if you're okay with merging this, warn me first so I can clean up the history.

PR #1975, opened 2020-01-19
Refactor Member Ignoring Method smell in SuggestionsProvider class

emaiannone:
Hi, I'm Emanuele Iannone, a master student at University of Salerno.
Since my bachelor's thesis I have been working on a **code smell refactoring plugin** called *aDoctor*, which is able to identify and fix energy-related problems in Android apps.
I launched it on your project, finding different instances of code smells. I chose one of them and let the plugin automatically fix it.
In this case I chose Member Ignoring Method, that is present when a non static method does not use at all instance variables and other non static methods. These kind of smell may have a non trivial impact on energy consumption, as shown in this paper: `https://www.sciencedirect.com/science/article/pii/S0950584918301678`.
Besides, this kind of refactoring does not impact on the functionalities of your app, so it is totally safe. Let me know if you are interested in this refactoring proposal.

### DIFF
--- a/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
+++ b/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
@@ -331,9 +331,9 @@ public class SuggestionsProvider {
         }
     }
 
-    private int calculateHashCodeForBuilders(List<DictionaryAddOnAndBuilder> dictionaryBuilders) {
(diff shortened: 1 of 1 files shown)


menny:
this is cool :)

menny:
I see that it used _tabs_ where my code uses _spaces_.

Also, it seems like this kind of change can be detected (and autofixed) with ErrorProne. Have you consider adding a plugin for ErrorProne?

PR #2052, opened 2020-02-14
Add German QWERTY keyboards

thomaseizinger:
This patch adds 3 keyboards to the German language addon, each of them offering a QWERTY variant of the existing layouts.

I am not an android developer, so I haven't been able to test this locally. I tried to install the output of the `build` task on my phone but it failed without a helpful error message.

Any pointers on where I can get a build from that I can install on my phone?
Is the CI going to produce artifacts that I can install?


menny:
it looks pretty good.
@martholomew what do you think?

@thomaseizinger 
1. connect your phone to your computer via usb.
2. run `adb devices`, you should see your device in the list.
3. run `./gradlew :addons:languages:german:apk:installDebug`. This will install the german pack onto your device.
4. verify.

thomaseizinger:
I just had an idea for another keyboard that might be useful, let me add that and test it locally :)

thomaseizinger:
Just pushed an updated version. I've added a "slim" variant that mimics the English Latin one :)

Tested it on my personal device and works as expected!
Good for review and merge from my side!

PR #2053, opened 2020-02-14
doc typo: Include :addons: in gradlew path.

rillian:
These example commands weren't updated when the language
repository was merged back into the main keyboard repo.


menny:
Thank you

PR #2056, opened 2020-02-15
Add Coptic language support

rillian:
Please consider adding this to the AnySoft package set. It's a bit rough, but I've been using it for a while and find it very useful for my own work in the language. I think it would benefit others to have it more widely available and would benefit in turn from feedback from other users.

[Coptic](https://en.wikipedia.org/wiki/Coptic_language) is the last stage of the ancient Egyptian language, written in a variant of the Greek alphabet. It's used liturgically as part of the Egyptian Christian tradition, by scholars studying the history of Egypt and Christianity, and as part of language revival efforts.

Current issues:
- [x] Passes tests.
- [ ] No diaresis, probably just long-press on vowels.
- [ ] Dictionary is pretty useless.
- [ ] Need to find a better way to support combining overbars.
- [ ] No stroke marks for the Boharic dialect.
- [ ] No double-hyphen for morphological analysis.
- [ ] No square bracket or combining underdot for manuscript annotation.

Some of these are easy, some I need help with. The dictionary in particular may be difficult. There's a small corpus of texts available under open licenses, but the existing segmenting scripts don't recognize Coptic words. What's included in the PR is based on a custom extraction I hacked up of various sources. Coptic has the same problem Greek does that a dictionary of all possible word forms is enormous, so it may always be limited without some kind of suggestion-filtering callback that knows some of the grammar.

There's also the issue of different dialects; there are two major ones. I don't know how useful it is to handle them as separate locales (with made up region codes??) or run everything together and let the user recognize the correct words.

Anyway, please let me know what you think.


addons/languages/coptic/apk/build.gradle
```diff
@@ -0,0 +1,7 @@
+ext.shouldBePublished = true
```
menny: Since this is a new pack, you should set this to `false`. There is a bit of a process to get packs published.

menny:
@martholomew do you have any comments?

addons/languages/coptic/apk/build.gradle
rillian: Thanks for taking a look. I've flipped the publication flag.

rillian:
Thanks!

PR #2073, opened 2020-02-22
Add middle dot and hyphen-minus as letter exceptions in Catalan

ccoreilly:
Catalan has many words using the middle dot (interpunct) between two L in order to represent the [ɫː] phoneme but when typing a middle dot in ASK it will stop suggesting those words as it does not consider it an inner word character. The same goes with the hyphen, used especially when inflecting verbs.

This PR adds both the middle dot and the hyphen to the IsLetter exceptions and the middle dot as a popup character of the hyphen.

### DIFF
--- a/addons/languages/catalan/pack/src/main/res/values/catalan_pack_strings.xml
+++ b/addons/languages/catalan/pack/src/main/res/values/catalan_pack_strings.xml
@@ -3,4 +3,5 @@
     <string name="catalan_dictionary_description">Diccionari de català</string>
     <string name="catalan_keyboard">Català</string>
     <string name="catalan_keyboard_description">Teclat QWERTY</string>
+    <string name="catalan_additional_is_letter_exceptions">&middot;&#45;</string>
 </resources>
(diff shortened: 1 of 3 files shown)


menny:
This is interesting. The dictionary has words with inner dots and hyphens? And your change works as you intended?

ccoreilly:
Yes, the dictionary has words like tranquil·la and tranquil·litat and after writing tranquil· suggestions are cleared.

However I was too quick assuming this would fix it as the additionalIsLetterExceptions are only considered in the ExternalAnyKeyboard nad not in the AnyKeyboard class. So no, this does not fix it, sorry.

I will look into it tonight and try to fix the behaviour.

ccoreilly:
Hi @menny,
I could debug why it was not working as I expected. the `additionalIsLetterExceptions` attribute must explicitly state the characters that are to be considered letter exceptions (referencing a @string made the Keyboard assume that the characters '@', 's', 't', 'r', and so one were the exceptions).

It now works as expected and writing `tranquil·` will not clear the suggestions.

ccoreilly:
I ran the app-tests-shards (0,2) checks locally using the same docker image (menny/ndk_ask:1.13.6) and they pass. Any idea what the problem might be?

ccoreilly:
I ran the app-tests-shards (0,2) checks locally using the same docker image (menny/ndk_ask:1.13.6) and they pass. Any idea what the problem might be?

menny:
It's a `SIGSEGV` issue with the JVM. The only option is to re-trigger this.
I can try to add an automatic re-try script, let me try that. In the meanwhile, I'll also re-trigger the checks.

Issue #1973, opened 2020-01-18
Improvements to Allow Indic Languages


martholomew:
Indic Languages from [The Indic Project](https://gitlab.com/indicproject/indic-keyboard) can be imported relatively easily; however, there are two main issues keeping this from happening:
- Shifted keys when they are combining characters are handled incorrectly #443 . This seems to be related to #1919 .
- An option to have the shift key only be manual toggle would assist in Indic Languages, Japanese and others (in other words, the shift key is not auto-toggled at the start of a sentence).

martholomew:
There is one more issue. Some Indic Languages have keys which when pressed produce a conglomeration of normal characters and combining characters. This functionality is not supported.
My solution would be for every character not "comma'd" off to be included in the output of a key. For example, right now you can comma off key codes for a cycle-through funcitonality:
`<Key android:codes="a,b,c" android:keyLabel="abc" android:popupKeyboard="@xml/popup_16keys_abc"/>`
So, the functionality could be changed so that if the keys were all next to each other without any commas, they would all be outputted. Of course, decimal codes could not be used for this as the boundaries are not obvious between them.
In the code it seems the key codes are stored in an array, so I will try to implement a 2D array for this purpose.

martholomew:
I implemented shiftedOutputText as a temporary measure, the 2D array is likely the best future solution to the issue, as it allows multi-character characters to be in popupkeys et cetera. Closed.

PR #2075, opened 2020-02-23
Fix Combining Characters when Shift Key is Toggled and Add AutoCap Var to Keyboard Files

martholomew:
This commit fixes half of #1973 , I'm working on figuring out how to solve the second half.
As you can see on my comment, the code I had to change doesn't do what it's supposed to, so it needs to be fixed a bit more than I have here.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
@@ -1348,10 +1348,14 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardColorizeNavBar {
     }
 
     private void updateShiftStateNow() {
+        boolean keyboardAutoCap;
+        AnyKeyboard alphabetKeyboard = getCurrentAlphabetKeyboard();
+        keyboardAutoCap = alphabetKeyboard.autoCap;
         final InputConnection ic = getCurrentInputConnection();
         EditorInfo ei = getCurrentInputEditorInfo();
         final int caps;
-        if (mAutoCap && ic != null && ei != null && ei.inputType != EditorInfo.TYPE_NULL) {
+        if (keyboardAutoCap && mAutoCap && ic != null && ei != null
+            && ei.inputType != EditorInfo.TYPE_NULL) {
             caps = ic.getCursorCapsMode(ei.inputType);
         } else {
             caps = 0;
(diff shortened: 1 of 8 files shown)


martholomew:
New commits fix the second part of #1973

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -1348,10 +1348,17 @@ public void onUpdateSelection(
     }
 
     private void updateShiftStateNow() {
+        boolean keyboardAutoCap;
```
menny: Instead of query for this every time this method is called, you can have it as a class member and set it inside `onAlphabetKeyboardSet()`

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
```diff
@@ -832,7 +832,13 @@ public AnyKey(
 
             // if the shift-character is a symbol, we only show it if the SHIFT is pressed,
             // not if the shift is active.
-            mShiftCodesAlways = mShiftedCodes.length == 0 || Character.isLetter(mShiftedCodes[0]);
+            // Note: above doesn't work, it only shows the symbols if SHIFT is being held.
```
menny: I'm also confused about the comment above....

Anyhow, what are you fixing here?

martholomew: If a shifted keycode is a combining character it doesn't show when shift is pressed.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
martholomew: Alright I moved it around!

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
martholomew: Basically the code block that i changed was supposed to make it so that unless shift was purposefully pressed, only alphabet letters would be shown. This stops special symbols from showing. I don't really understand the functionality except on bottom bars where you can have an exclamation mark that changes into a question mark or something on shift press. But, this code doesn't even work.

martholomew:
I've added the unit tests, but it seems that the unit tests overall which use xml are having issues right now.

menny:
I've created a PR based on yours: https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2078

There are two commits on top of yours which fix the tests (there was no error in your code).

menny:
You can either close this PR and merge the other one (the commits are still under your name), or you can cherry-pick the extra two commits from the other PR and push them into this PR.

martholomew:
Awesome, sounds good to me.

PR #2081, opened 2020-02-27
ASK: Added shiftedOutputText attr, auto label them; Sinhala: Added

martholomew:
Sorry, there are a couple of changes here:

Keyboard, PointerTracker: Added the attribute shiftedOutputText to be able to output multi-character "strings" when shifted.

AnyKeyboardViewBase: Made it so that "shiftedOutputText" keys are auto label'd so that you don't have to add a shiftedKeyLabel attribute.

AnyKeyboard: Changed mShiftCodesAlways to true so that it doesn't have any problems with [ZWNJ](https://en.wikipedia.org/wiki/Zero-width_non-joiner) and [ZWJ](https://en.wikipedia.org/wiki/Zero-width_joiner) keys which are used in Indic languages.

Sinhala: Added the language as it uses the new features.


martholomew:
There is one hanging issue. Since shiftedOutputText uses the onText function in AnySoftKeyboardSuggestions, suggestions are restarted whenever one is entered. I'm not sure if I should look at changing the functionality or if I should make a seperate function. Since this is a bit of an edge case, I'm sure this can be merged without worrying about it, as this also occurs [here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/addons/languages/hebrew/pack/src/main/res/xml/heb_qwerty_niqqud_shin.xml).

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
```diff
@@ -833,12 +833,7 @@ public AnyKey(
             // if the shift-character is a symbol, we only show it if the SHIFT is pressed,
             // not if the shift is active.
             // Note: above doesn't work, it only shows the symbols if SHIFT is being held.
-            mShiftCodesAlways =
-                    mShiftedCodes.length == 0
-                            || Character.isLetter(mShiftedCodes[0])
-                            || Character.getType(mShiftedCodes[0]) == Character.NON_SPACING_MARK
-                            || Character.getType(mShiftedCodes[0])
-                                    == Character.COMBINING_SPACING_MARK;
+            mShiftCodesAlways = true;
```
menny: does this mean that `mShiftCodesAlways` is `true` is all cases? Can we remove it and all the branches that use it?

martholomew: Sounds about right.

martholomew:
I don't think we need new unit tests.

menny:
Good. Then please remove that field and related code. This is great - I love code deletion

menny:
thanks!

martholomew:
Awesome, no prob!

PR #2092, opened 2020-03-02
adnw: some capital keys fixed

chrbauer:


menny:
@chrbauer can you explain this change? I'm not sure I understand why the shifted character for `f` is `ẞ`.
@martholomew can you review this?

chrbauer:
This was the first keybord layout I made. So I started from the neo2 layout, because adnw shares all the special keys from neo2. Unfortunately I overlooked the shiftCodes. So in the current version shift f is ẞ. After my patch it is F.

PR #2097, opened 2020-03-05
Fixed onText issues with auto correct

### LINKED ISSUE #2096: "shiftedOutputText" and "outputText" Reset Word Prediction
This behavior can be observed on the "shiftedOutputText" keys on the Sinhala Inscript keyboard. On being pressed, the word predictions are reset.

This is because it uses the onText function which is supposed to have this behavior.

There are three solutions:

- Change the onText function's behavior, which will inadvertantly change quickText behavior.

- Make a new function that the outputText keys will use.

- Get rid of shiftedOutputText and switch the keycodes to using a 2d array so that all the keys can have multi-character outputs, each comma'd off from the other.

The second solution will be the easiest immediately, while the third makes more sense in the long run.

martholomew:
Fixes #2096


ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
```diff
@@ -466,15 +466,41 @@ private void detectAndSendKey(int index, int x, int y, long eventTime) {
...
+                         * debouncing is in effect.
+                         */
+                        if (nearByKeyCodes.length >= 2
+                                && nearByKeyCodes[0] != code
+                                && nearByKeyCodes[1] == code) {
+                            nearByKeyCodes[1] = nearByKeyCodes[0];
+                            nearByKeyCodes[0] = code;
+                        }
+                        if (listener != null) {
+                            if (isInGestureTyping()) {
```
menny: I'm not sure I understand this part. Can you add some comments?

ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
```diff
@@ -466,15 +466,41 @@ private void detectAndSendKey(int index, int x, int y, long eventTime) {
...
-                if (listener != null) {
-                    listener.onText(key, key.shiftedText);
-                    listener.onRelease(0); // dummy key code
+                    CharSequence text = null;
+                    if (!mKeyDetector.isKeyShifted(key)) {
+                        text = key.text;
+                    } else if (mKeyDetector.isKeyShifted(key)) {
+                        text = key.shiftedText;
+                    }
+                    int[] nearByKeyCodes = mKeyDetector.newCodeArray();
```
menny: I was wondering about the near-by-codes when you created the issue.
Maybe we do not need that at all? Maybe we can provide a single (new) array here with just the code? These near-by-codes are used mostly to generate corrections due to user pressing a neighbor key instead of the key they actually wanted.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
martholomew: I'm afraid I'm not 100% on it either, as I just copied and reordered the code a little further down.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
martholomew: Would we get rid of that entirely, then?

menny: yes, I think so..

ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
menny: we need some unit-tests here so we'll know how it behaves.

martholomew:
So, next on this, we'll be getting rid of the nearby codes? Was there anything else that needed to be cleared up?

ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
```diff
@@ -460,21 +460,48 @@ private boolean shouldLongPressQuickly(Key key) {
...
+                    if (!isShifted) {
+                        text = key.text;
+                    } else if (isShifted) {
+                        text = key.shiftedText;
+                    }
+                    int[] nearByKeyCodes = mKeyDetector.newCodeArray();
+                    mKeyDetector.getKeyIndexAndNearbyCodes(x, y, nearByKeyCodes);
+                    mTapCount = 0;
+                    for (int i = 0; i < text.length(); i++) {
+                        int code = text.charAt(i);
```
menny: after this line, you can create the `nearByKeyCodes` array:
```
nearByKeyCodes = new int[] { code };
```

Right?

ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java
menny: I think this should be called once, before the loop. Again, I'm not really sure what this code does...

menny:
@martholomew any update on this?

martholomew:
Been working on Telex, lol.

menny:
@martholomew I've made some changes to your branch. I will see if that works, but I need your help with test-cases. Can you give me some examples of what `text` and `shiftedText` to verify?

martholomew:
Hey @menny , sorry I've been slow on this. I'm not sure when I'll be able to get more coding done, so thanks for the help on this one.
`text` should output a string of characters
`shiftedText` should output a string of characters when the shift key is engaged
Is that what you needed to know?

menny:
You have tried to fix https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/2096. Can you give me some text examples where the actual keyboard behavior was wrong?

menny:
@martholomew I'm doing some serious refactoring here....

PR #2098, opened 2020-03-05
Coptic Fixes

martholomew:
@rillian I've added a new Coptic keyboard based on ~~swiftkey's~~ with a key that adds the macron or the grave accent based on how many times you tap the key. If you could test it I would be grateful.

In regards to the dictionary, I've got a hacky script I've used with apparent success for other languages. Would you be able to point me to a corpus I could give a crack at?

Coming back to the keyboards. The scriptorium keyboard could fit, you would just need to move one alphabet key from the third row to the second. I don't know coptic so I couldn't say which would be best to move, so your input would be helpful!

I'll copy your list over:
- [ ] No diaresis, probably just long-press on vowels.
- [ ] Need to find a better way to support combining overbars.
- [ ] No stroke marks for the Boharic dialect.
- [ ] No double-hyphen for morphological analysis.
- [ ] No square bracket or combining underdot for manuscript annotation.

Could you inform me on how you would like to see these implemented?


addons/languages/coptic/pack/src/main/res/xml/coptic_scriptorium.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
-    android:keyWidth="10%p"><!-- "android:keyWidth" specify the default width of a key. In this example, 10% of parent (the whole keyboard) -->
+    android:keyWidth="9.09%p"><!-- "android:keyWidth" specify the default width of a key. In this example, 10% of parent (the whole keyboard) -->
```
rillian: Not sure about changing the key width, I already have trouble getting the correct key. I'll give it a try though.

Anyway, the comment should be updated to match, maybe just to say that 10% is the usual width.

addons/languages/coptic/pack/src/main/res/xml/coptic_swift.xml
```diff
@@ -0,0 +1,37 @@
+<Keyboard xmlns:android="http://schemas.android.com/apk/res/android" xmlns:ask="http://schemas.android.com/apk/res-auto" android:keyWidth="10.0%p" android:keyHeight="@integer/key_normal_height" ask:autoCap="false">
+  <Row>
+    <Key android:keyEdgeFlags="left" android:codes="ⲑ"/>
+    <Key android:codes="ϣ"/>
```
rillian: Wow! I had no idea you could give literals for `android:codes`. That's *so* much easier than decimal codepoints!

addons/languages/coptic/pack/src/main/res/xml/coptic_swift.xml
```diff
@@ -0,0 +1,37 @@
...
+    <Key android:keyEdgeFlags="right" android:codes="ⲡ"/>
+  </Row>
+  <Row>
+    <Key android:keyEdgeFlags="left" android:codes="ⲁ"/>
+    <Key android:codes="ⲥ" android:popupCharacters="ⲋ"/>
+    <Key android:codes="ⲇ" android:popupCharacters="⳧"/>
+    <Key android:codes="ϥ" android:popupCharacters="ⲫ"/>
+    <Key android:codes="ⲅ"/>
+    <Key android:codes="ⲏ" android:popupCharacters="ϩ"/>
+    <Key android:codes="ⲓ" android:popupCharacters="ϭ"/>
```
rillian: This duplicates the iota in the upper row. Probably should be ϫ which is otherwise missing.

rillian:
Hi @martholomew! I'll give the new layouts a try tomorrow and offer what usability impressions I can.

In the meantime, some quick responses to some of your questions:

- [Public corpus](https://github.com/CopticScriptorium/corpora) of edited Coptic texts
- There are also a bunch of 19th/early 20th text editions at archive.org, if you want to try your hand at OCR. This [Pistis Sophia](https://archive.org/stream/CarlSchmidtPistisSophia1925Teil1/Carl%20Schmidt%2C%20Pistis%20Sophia%201925%2C%20Teil2#mode/1up) for example.
- [Dictionary data](https://github.com/KELLIA/dictionary)
- Having a popup ⲓ̈ and maybe ⲩ̈ should take care of the diaresis.

I'm not sure what you mean about the scriptorium layout fitting. It has 32 letters, laid out in rows of 11/11/10 keys, plus the combining macron. I don't see how that fits the AnySoftKeyboard template of 10/9 or 10/7 keys.

To be clear, I find using popups fine, it's just another source of variation between layouts, and I'm not sure how to decide what to move to popups: logical, position in roman or greek layouts, frequency of use? All of which is made worse by not having a working autocorrect dictionary.

Anyway, I'll try out your changes and give you more feedback in a day or two.

addons/languages/coptic/pack/src/main/res/values-fr/strings.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="keyboard_name">ⲕⲏⲙⲉ</string>
+    <string name="keyboard_name_scriptorium">ⲕⲏⲙⲉ Scriptorium</string>
+    <string name="keyboard_name_swift">ⲕⲏⲙⲉ Swift</string>
```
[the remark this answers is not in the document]
rillian: I take it back. I still like the ⲕⲏⲙⲉ, at least in the noto font.

One idea for shortening this would be to use **ⲕⲏⲙⲉ ͼ**, replacing _scriptorium_ with GREEK DOTTED LUNATE SIGMA SYMBOL, the editorial mark the project uses for a logo.

Swift might be something like **ⲕⲏⲙⲉ ⲙ̅ⲁⲥⲁⲓ**.

addons/languages/coptic/pack/src/main/res/xml/coptic_swift.xml
martholomew: For sure, but it does get confusing with combining characters.

addons/languages/coptic/pack/src/main/res/xml/coptic_swift.xml
martholomew: I messed up there, fixed it.

martholomew:
I made a dictionary based on the public corpus. There might be something weird about the diacritics, but I'm not sure. Tell me how it works!

For the coptic "gamma/Y" and "iota", I ended up using a diaresis y and diaresis iota without combining characters as a temporary measure. Popup characters can't be more than one unicode character, or it will split them up, i.e., show a y and next to it the diaresis combining character. That's something that needs to be fixed anyway, so I'll see about doing that.

addons/languages/coptic/pack/src/main/res/xml/coptic_swift.xml
```diff
@@ -6,8 +6,8 @@
     <Key android:codes="ⲣ" android:popupCharacters="ⳁ"/>
     <Key android:codes="ⲧ" android:popupCharacters="ϯ"/>
     <Key android:codes="ⲯ"/>
-    <Key android:codes="ⲩ"/>
-    <Key android:codes="ⲓ"/>
+    <Key android:codes="ⲩ" android:popupCharacters="ÿ"/>
+    <Key android:codes="ⲓ" android:popupCharacters="ϊ"/>
```
rillian: Popups should be ⲩ̈ and ⲓ̈. (Coptic letter ua and copic letter iauda.)

rillian:
> Popup characters can't be more than one unicode character, or it will split them up, i.e., show a y and next to it the diaresis combining character.

Aha. Thanks for clarifying!

rillian:
The dictionary is better than the one I did. It's still not much help correcting typos, but it doesn't actively sabotage things either. :)

I haven't checked the diacritics yet.

martholomew:
Alright, let me know.

rillian:
I looked at the diacritics in the dictionary a bit. They're all over the place, as might be expected since the source data is as well. :smile:

Running [this quick script](https://gist.github.com/rillian/7e06def080c74df26bbd9a3efafa8d25) to count combining marks:
```
u+000304 COMBINING MACRON 513
u+000305 COMBINING OVERLINE 663
u+00fe24 COMBINING MACRON LEFT HALF 497
u+00fe25 COMBINING MACRON RIGHT HALF 461
u+000307 COMBINING DOT ABOVE 987
u+000308 COMBINING DIAERESIS 280
u+00fe26 COMBINING CONJOINING MACRON 103
u+000302 COMBINING CIRCUMFLEX ACCENT 181
u+000323 COMBINING DOT BELOW 469
u+002cef COPTIC COMBINING NI ABOVE 31
u+002cf1 COPTIC COMBINING SPIRITUS LENIS 2
u+000300 COMBINING GRAVE ACCENT 3
```
For a dictionary these should probably be normalized a bit.
- `COMBINING DOT BELOW` is used as a editorial mark to indicate uncertain readings in manuscript transmission. It can just be stripped; any word appearing in the corpus is by definition considered attested by an editor.
- `COMBINING MACRON` is the default syllabic consonant mark for the Sahidic dialect. It should stay.
- `COMBINING OVERDOT` is the corresponding mark for Boharic. `COMBINING GRAVE ACCENT` is the same as the overdot. The unicode spec ([section 7.3](https://www.unicode.org/versions/Unicode13.0.0/ch07.pdf)) says either is acceptable. In the dictionary one should be normalized to the other. Your swiftkey layout uses grave, which I guess would be more familiar to users coming from modern Greek. On the other hand, more recent editors seem to prefer the overdot, at least those publishing in English. Compare the relative frequencies in your dictionary: 987 overdots to 3 graves.
- `COMBINING OVERLINE` is used to join the letters of common abbreviations. So is the combination `COMBINING MACRON HALF LEFT`, `COMBINING CONJOINING MACRON`, `COMBINING MACRON HALF RIGHT`. The later often looks a little nicer, but of course requires input method support to generate conveniently. They seem about equally popular in your extracted word list. Maybe we can decide based on what looks better in the current Noto font?
- `COMBINING OVERLINE` is sometimes used over letters to indicate they are to be read as numbers. Unicode specifies this character in particular, along with `COMBINING DOUBLE OVERLINE` and `GREEK LOWER NUMERAL SIGN`, so if we normalize away from this sign we should make sure not to do it to the numbers.

The other marks are rare; I think we can just leave them as-is.

Would you be willing to contribute your dictionary script to the repository so we can document any normalization and make it easier to update the dictionary when new texts are published?

martholomew:
@rillian Awesome.

[I put together what I have into a repository](https://github.com/martholomew/make-dic). Uses bash and perl, hopefully my instructions are good enough.

menny:
So, what's going on with this PR?
Also, if you want this to be published, please make sure all the required information (including screenshots) is correct in the `play` folder

martholomew:
@menny I'm just waiting for some more feedback from @rillian , I believe.

rillian:
Sorry, got distracted.

- This is an improvement so I think it ought to be merged.
- I find the double-tap for overbar annoying, so it would be good to have a Sahidic variant with the order switched.

Thanks for posting your dictionary scripts! I haven't had a chance to work out if they're doing something reasonable, but as I said, the results are better than my attempt.

I did notice that your copcorp.txt seems to be broken into morphs, rather than words. I think that's causing problems e.g. ⲣⲱⲙⲉ is in your dictionary but not the more common ⲡⲣⲱⲙⲉ with the definite article prefix. Probably we need to read the annotation data to re-assemble the word boundaries.

I also thought you were writing a PR to address the combining-diacritic+letter issue with the longpress diaresis. How did that go?

martholomew:
I see, coptic is a highly inflected language like ancient greek? I'm not sure how I would edit the dictionary script for that, I guess it could be spell checked? Tell me what you think.
In regards to the longpress issue, that requires a bit of change, I'll get to that once I can.

rillian:
That's right. Which means an effective dictionary based on static matching will be huge. As a dead language there's also the problem of missing attestations for a lot of possible forms.

However, I think the immediate problem we have is that the texts you used were sometimes broken into morphemes instead of words. I'll see if I can come up with something better.

menny:
any news about this?

rillian:
I've not had a chance to improve the dictionary scripts.

I generally find the swift-layout good, and having had a chance to use it for the Bohairic dialect last week, it seems better suited to that. double-tapping to get the Sahidic overbar is tedious. Perhaps it would be better to use it specifically for that dialect?

The scriptorium layout may be easy for beginners, but personally I find it unusable without a backspace key. @martholomew, what did you think of removing the shift key instead? Upper case is more typographic than grammatical for Coptic.

rillian:
@menny How are dialect or regional spelling differences handled in AnySoftKeyboard? Is it possible to switch the dictionary with the layout within a language add-on, or must they be packaged as separate languages?

addons/languages/coptic/pack/src/main/res/xml/coptic_scriptorium.xml
```diff
@@ -55,6 +54,6 @@
         <Key android:codes="1005"/>
         <Key android:codes="1007"/>
         <Key android:codes="11407"/>
-        <Key android:keyWidth="15%p" android:codes="-5" android:keyEdgeFlags="right" android:isRepeatable="true"/>
+        <Key android:codes="-5" android:keyEdgeFlags="right" android:isRepeatable="true"/>
```
rillian, line 57: I'm confused what you meant by moving a key from the third row to the second; the second row seems full already?

If you want to move a key rather than dropping the shift, I would suggest making `ⳉ` (11465) a popup on `ϧ` (999), and then moving `ⲏ` (11407) up to the right edge of the second row. ⳉ has the same value as ϧ and is used only in the less common Akhmimic dialect, so it's probably the easiest to lose. ⲏ is quite common, so moving it up a row is reasonable, while keeping the position conceptually similar.

menny:
A pack can have multiple dictionaries. Each dictionary exposing its locale id. Layouts can declare what is the default locale to be used

rillian:
Ok, that's great. Thanks!

menny:
FYI, Coptic was released to Google Play Store: https://play.google.com/store/apps/details?id=com.anysoftkeyboard.languagepack.coptic

rillian:
> FYI, Coptic was released to Google Play Store

Thanks, that's great news! Five installs! :)

Are you planning to release it on F-Droid as well?

rillian:
BTW, Gboard apparently [added Coptic](https://support.google.com/gboard/answer/6380730?visit_id=637485958589433981-1316472083&rd=1#languages&zippy=%2Csee-supported-languages) recently. Yet another layout!
gboard-coptic-layout Thanks to Ⲛⲁⲑⲁⲛ for the screenshots.

The language name is in the Bohairic dialect. They seem to have all the diacritics covered, even if the display is a little ugly. I'm not clear if there's any prediction.

Issue #1875, opened 2019-11-12
Fast typing moves keyboard cursor back a character most of the times


RenatoLopes771:
It's so annoying. It makes me intentionally type slow so the thing doesn't make this kind of behaviour. Please fix.

(What I mean by the title is it would be the equivalent of pressing that arrow key at the top withouth finishing typing a word)

ArenaL5:
Duplicate of #1290.

ArenaL5:
I can verify this bug, and I believe I've found the conditions that trigger it. If I'm right, I'd really like you to check this bug after PR #1893 is corrected and merged, as I just removed a function which might have bene making your bug worse.

For me, this bug happens only if:
- word prediction is enabled, and the word I'm typing is underlined.
- I press on two keys (let's say, first A, then S) before releasing any.
- I release key S before releasing key A, outputting "as" but making the cursor advance only one letter instead of two.

If someone can verify or deny these conditions, it would greatly help to eventually fix this bug. 

(For the moment, if you want to keep using AnySoftKeyboard while this bug is not fixed, try closing the suggestions bar or disabling suggestions altogether.)

YoshK:
Hey,
I have the same issue on a Huawei P20 pro 9.1
Any workaround has been found?

menny:
Are you seeing this with the latest release?

YoshK:
The last release accessible through F-Droid at least: 1.10.1096

dkathmann:
> Are you seeing this with the latest release?

I'm on 1.10.1096 (Play Store) and the bug still occurs.

karpad3:
I'm on 1.10.1096 as well and the bug still occurs. It's like that in the last few months.
I'm using Xiaomi Redmi Note 5 with Android 9.

menny:
I can not reproduce this, but I have other devices. Could you provide a video capture?

dkathmann:
@menny [video capture download](https://send.firefox.com/download/fe56c7118b3f328c/#D4Q2JF4HCQ46M6Bmm6bokw)

technOdin:
I have the same issue on my Huawei P20 Pro.

menny:
@dkathmann the link expired. Can you send again?

dkathmann:
@menny [reupload](https://send.firefox.com/download/3bfa9675e8861851/#FTagEmbPKbn1uqC9vKZSSQ)

vocux:
[...]

PR #2107, opened 2020-03-08
Fix cursor lagging (#1875)

ArenaL5:
The main commit in this pull request is b18bd87fa58467791ac07029fd6d9cb47d028bd7. It updates the global cursor position after every character is handled, which has normally no effect (as it's overwritten when `onUpdateSelection` is called), but takes into account the effects of every key when several keys are pressed, not just the last one.

This should fix bug #1875.

The last commit is the one I'm the most unsure about: that test is completely unrelated to my fix, but it keeps failing in my local master branch.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
@@ -621,14 +621,13 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
                     cursorPosition = -1;
                 }
 
-                /* mGlobalCursorPosition is updated when onUpdateSelection is called,
-                 * but that doesn't happen if several keys are pressed simultaneously */
-                mGlobalCursorPosition = cursorPosition;
-
                 ic.setComposingText(mWord.getTypedWord(), 1);
                 if (cursorPosition > 0) {
                     ic.setSelection(cursorPosition, cursorPosition);
                     ic.endBatchEdit();
+                    /* mGlobalCursorPosition is updated when onUpdateSelection is called,
+                     * but that doesn't happen if several keys are pressed simultaneously */
+                    mGlobalCursorPosition = cursorPosition;
                 }
             }
             // this should be done ONLY if the key is a letter, and not a inner
--- a/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
@@ -621,6 +621,10 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
                     cursorPosition = -1;
                 }
 
+                /* mGlobalCursorPosition is updated when onUpdateSelection is called,
+                 * but that doesn't happen if several keys are pressed simultaneously */
+                mGlobalCursorPosition = cursorPosition;
+
                 ic.setComposingText(mWord.getTypedWord(), 1);
                 if (cursorPosition > 0) {
                     ic.setSelection(cursorPosition, cursorPosition);
--- a/scripts/adb.sh
+++ b/scripts/adb.sh
@@ -38,7 +38,7 @@ else
             exit 1
         fi
     done
-    tags=`grep -R 'TAG = ".*"' app/src/main/java/com/* jnidictionaryv1/src/main/java/com/* jnidictionaryv2/src/main/java/com/*`
+    tags=`grep -R 'TAG = ".*"' ime/app/src/main/java/com/* ime/jnidictionaryv1/src/main/java/com/* ime/jnidictionaryv2/src/main/java/com/*`
     # We can go back to our original folder now:
     cd "$oldpath"
     tags="$(echo $tags | sed -E 's![a-z/A-Z12]*\.java: (protected |private )?(static )?(final )?String [A-Z_]* = "([^\"]*)";!\4!g')"


ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
```diff
@@ -621,6 +621,10 @@ protected void handleCharacter(
                     cursorPosition = -1;
```
menny: This seems problematic. maybe it should be set to `mGlobalCursorPosition`?

menny:
I'm also find the need to update that test concerning. Can you reproduce it on a real device?

ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
ArenaL5: I'm not too sure about that... `cursorPosition` is made negative on purpose because the code works differently at the end of a word than at the middle. Presumably it's faster at the end, I didn't write that part.

(**BTW!** I pushed another commit to move my fix somewhere else, because I mistakenly overwrote `mGlobalCursorPosition` to -1 sometimes. It surprisingly didn't change anything because the code depending on that variable had no chance to run then, but this way it's written to less times and raises fewer questions.)

menny:
@ArenaL5 can you write a test for this fix? It seems very fragile case that can easily regress. This is too important case for allowing regression

ArenaL5:
> I'm also find the need to update that test concerning. Can you reproduce it on a real device?

[This ZIP](https://github.com/AnySoftKeyboard/AnySoftKeyboard/files/4303888/AnySoftKeyboardGimmicksTest.zip) has a MP4 screen grab while pressing the apostrophe and Shift keys manually, following the code. Does that work?

> can you write a test for this fix?

I'll do it, but I may need some time to grasp how tests work here. (I can't run forever from tests, can I?)

menny:
> [This ZIP](https://github.com/AnySoftKeyboard/AnySoftKeyboard/files/4303888/AnySoftKeyboardGimmicksTest.zip) has a MP4 screen grab while pressing the apostrophe and Shift keys manually, following the code. Does that work?
> 
okay, then the original test is correct. You should not change the test in your PR. Something in your test broke the required behavior (only holding SHIFT should produce `"`, in all other cases - caps-lock or shifted-but-not-held - it should produce `'`)

> 
> I'll do it, but I may need some time to grasp how tests work here. (I can't run forever from tests, can I?)

How can I help with that? I run tests from the command-line in most cases.
If I want to run a specific class:
```
./gradlew :ime:app:testDebugUnitTest --tests="*AnySoftKeyboardGimmicksTest*"
```

ArenaL5:
Okay. I did a force-push to undo my changes to `AnySoftKeyboardGimmicksTest.java`.

For the new test, I'm running into a different problem than I expected: I'm adding this code to the bottom of `AnySoftKeyboardSuggestionsTest.java`:

```java
    @Test
    public void testCursorSyncsUpOnSimultaneousKeypresses() {
        Assert.assertEquals(
                "", getCurrentTestInputConnection().getCurrentTextInInputConnection());

        mAnySoftKeyboardUnderTest.simulateKeyPress('l');
        mAnySoftKeyboardUnderTest.getCurrentInputConnection().setSelection(0, 0);
        mAnySoftKeyboardUnderTest.onPress('h');
        mAnySoftKeyboardUnderTest.onPress('e');
        mAnySoftKeyboardUnderTest.onRelease('e');
        mAnySoftKeyboardUnderTest.onRelease('h');
        Assert.assertEquals(
                "hel", mAnySoftKeyboardUnderTest.mWord.getTypedWord().toString());
    }
```

The test builds and runs, but the assertion fails because the H and the E are never added. I'm clearly using the wrong methods from `mAnySoftKeyboardUnderTest` but I don't know what's the code for "hold key H, hold key E, relase key E, release key H".

menny:
can you try:
```
    @Test
    public void testCursorSyncsUpOnSimultaneousKeypresses() {
        Assert.assertEquals(
                "", getCurrentTestInputConnection().getCurrentTextInInputConnection());

        mAnySoftKeyboardUnderTest.simulateKeyPress('l');
        mAnySoftKeyboardUnderTest.getCurrentInputConnection().setSelection(0, 0);
        mAnySoftKeyboardUnderTest.onPress('h');
        mAnySoftKeyboardUnderTest.onPress('e');
        mAnySoftKeyboardUnderTest.onKey((int)'e', .....);
        mAnySoftKeyboardUnderTest.onRelease('e');
        mAnySoftKeyboardUnderTest.onKey((int)'h', .....);
        mAnySoftKeyboardUnderTest.onRelease('h');
        Assert.assertEquals(
                "hel", mAnySoftKeyboardUnderTest.mWord.getTypedWord().toString());
    }
```
I look at https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/PointerTracker.java#L509 and it seems the `onKey` is called before `onRelease`
You should simulate something like that. Also, I'm not really sure about the arguments to `onKey`. Try to look at [findKeyWithPrimaryKeyCode](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/test/java/com/anysoftkeyboard/TestableAnySoftKeyboard.java#L398) or [simulateKeyPress](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/test/java/com/anysoftkeyboard/TestableAnySoftKeyboard.java#L348) for ideas how to use that.

ArenaL5:
Sorry for taking so long, I was very tired this couple of days and didn't really get to try this extensively until now.

It doesn't seem to be working. Sure, with `onKey` keys are registered (I can type "hel"), but they're registered in the same order even without my fix. It seems like the keys are still typed one at a time instead of using multitouch.

I also tried copypasting code from AnySoftKeyboardGimmicsText.java, but it stopped working when I changed the Shift key with a letter key.

menny:
Probably we don't really understand what's going on there...
if you can reproduce this issue on your device, then I suggest adding logging to all the relevant methods and after reproducing the issue, go over the log and see what is the actual process

menny:
okay, I tried out my suggestion, and it didn't work.
But, I tried a different approach and found an unrelated [issue](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2121).

I also added a unit-test that simulate the issue you reported, but from the View interaction: https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2121/files#diff-489990b0fd511923bf6a80199f2867f7R404

menny:
but, my unit-test does not simulate or fix the issue reported at #1875. Your fix might.

menny:
I've added your fix here: https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2137

ArenaL5:
Perfect. I was having problems with programming a working test for it, I'll check later how you did it.

Thank you very much. :relaxed: I've read the bug still exists, but it's not as bad. I suppose it only happens with three or more simultaneous fingers.

menny:
I think my solution only handles the case where there is one delayed OS event. It should be expended, perhaps.. not sure.

PR #2137, opened 2020-03-16
Fix lagging

menny:


martholomew:
Ah, so this is needed, then?

menny:
yes, seems so... the unit-tests failure condition was broken. I'm very sad about that.

menny:
@martholomew can you review, please?

PR #2150, opened 2020-03-18
Italian layouts

### LINKED ISSUE #1736: Pointless keys in italian layout
### Steps to reproduce
Italian keyboard has a lot of useless symbols and keys.

### Expected behaviour
There should be parentesys, apostrophe (it could stay at the Lʾs right as in the English layout, or in the long press of "c"), quotation marks, euro symbol. 

### Actual behaviour
There are plenty of useless symbols. Italian has accents only on vocals. Consonants with accents are useless. See attachment for how an italian keyboard layout should be. Plus, giving the opportunity to customize every key would be great. It would mean less work for you 

**Android OS version:**
9 pie. On Tugapower 17 custom rom

Oneplus One

Italian, English 
Screenshot_20190603-031712_Browser

nicoursi:
Fixes #1736

Created an Italian with symbols layout.
I also added popup domains to the it locale


addons/languages/italian/pack/src/main/res/xml/italian_autotext.xml
```diff
@@ -20,6 +20,25 @@
     <!-- this dictionary is used for common typing mistakes -->
     <!-- the attribute 'src' will be replace by the node's value: -->
     <!-- rght will be converted to right -->
-    <word src="rght">right</word>
-    <word src="riht">right</word>
+
+
+    <!-- autotext needs to be improved. e.g if the user types (after a dot) Cone, it will be replaced by come and not Come  -->
+
+    <word src="cone">come</word>
```
menny: this is great. thanks

addons/languages/italian/pack/src/main/res/xml/italian_qwerty_with_symbols.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
```
menny: I want to give @martholomew the option to review this.

nicoursi:
Can we get rid of Italian_querty?
If so, what name should we keep? Italian_querty or italian_querty_with_symbols?

menny:
can you add detail about this change to https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/addons/languages/italian/apk/src/main/play/release-notes/en-US/alpha.txt

nicoursi:
> can you add detail about this change to https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/addons/languages/italian/apk/src/main/play/release-notes/en-US/alpha.txt

done

nicoursi:
I added another important commit. In italian we don't have the word `po`.  We hve the word `poco` that gets truncated and becomes `po'`.
In the dictionary `aosp.combined.gz` there is the following entry:

` word=po,f=136,flags=,originalFreq=136`

Whenever the user wants to write `po'`, he'll only get suggestions for `po`, even if `po` doesn't really exist, and it's considered a mistake in italian.

Commit #2420b96 fixes this by replacing the entry above with the following one:

`word=po',f=136,flags=,originalFreq=136`

It's tested and working. I also added this change in release notes [release notes](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/addons/languages/italian/apk/src/main/play/release-notes/en-US/alpha.txt)

nicoursi:
@menny 

I added another important commit. In italian we don't have the word po. We hve the word poco that gets truncated and becomes po'.
In the dictionary aosp.combined.gz there is the following entry:

word=po,f=136,flags=,originalFreq=136

Whenever the user wants to write po', he'll only get suggestions for po, even if po doesn't really exist, and it's considered a mistake in italian.

Commit #2420b96 fixes this by replacing the entry above with the following one:

word=po',f=136,flags=,originalFreq=136

It's tested and working. I also added this change in release notes release notes

menny:
Thanks! I'm having difficulties finding time to review PRs at the moment. This looks like an important PR. Sorry for the delay

PR #2159, opened 2020-03-20
Spanish locale

### LINKED ISSUE #2074: [Button Change] [Spanish Layout] Change "¿" and "¡" characters in symbols keyboard [Low priority]
This is a small change in the Spanish layout because in Spanish the characters "¿" and "¡" are more used than other more easily accesed with the actual layout.

This are the actual layout
ASK_Spa_SymbolsOriginal

And this is what make more sense in Spanish
ASK_Spa_SymbolsFixed

I know you can access all these characters in the "letters" keyboard pressing the "." button, but it's a detail to make more sense with the keyboard.

nicoursi:
Fixes #2074 

Commit [7185d6](https://github.com/AnySoftKeyboard/AnySoftKeyboard/commit/7185d6e060cb9db519df0f506927a52f8e160abe) fixes the currency issue.
Commit [07558cd](https://github.com/AnySoftKeyboard/AnySoftKeyboard/commit/07558cd2499667e2d83360edffa63c3577632c76) could be improved. 

I think it should not be bound to the locale but to the language addon being used.

If I am italian, and my phone language is set to Italian, when I type in Spanish, I cannot benefit from the easily accessible `¿` and `¡`. These symbols are bound the the language you are writing into, and not to the one set as default language in Android.

I tried to move a modified version of [symbols.xml](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/res/xml/symbols.xml) to [xml](https://github.com/AnySoftKeyboard/AnySoftKeyboard/tree/master/addons/languages/spain/pack/src/main/res/xml), but it does not work.

If there's a way to do it, I'll be happy to give it a try.

### DIFF
--- a/ime/app/src/main/res/values-es/currency_data.xml
+++ b/ime/app/src/main/res/values-es/currency_data.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <integer name="default_currency_sign_unicode">8364</integer>
+    <string name="default_currency_sign">&#8364;</string>
+    <string name="popup_currency_signs">$₽¥¢&#8377;£&#162;₢€₿</string>
+</resources>
--- a/ime/app/src/main/res/xml-es/symbols.xml
+++ b/ime/app/src/main/res/xml-es/symbols.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+
+<Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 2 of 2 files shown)


dilworks:
Sorry for commenting on an already merged PR, but not all Spanish locales use the Euro as their currency.

Particularly we at Latin America we mostly use the $ (peso, dollar, etc.), so having the € as the default currency key is annoying.

Is there a way to fix/customize this?

menny:
This is a good point.
It is fixable with extended locales.
Please open a new issue with the details

PR #2167, opened 2020-03-22
added popupcaracters to de_querty and de_quertz

### LINKED ISSUE #1925: Feature request: In german language is missing the § (for paragraph)
I can't find the sign § in the german language pack for android.

It could be as popup behind the "s".

### LINKED ISSUE #2010: German: Add "German+" Layout
Similar to how there is a "German Broad" and a "German Broad +" layout, there should be a keyboard with extra pop-ups that uses the standard "German" Layout. (the one with ä,ö, and ü on the right side)

Ideally, it would be the same pop-ups that "German Broad +" - layout uses... while still keeping the special letters on the right the same.

(This might be the same thought this feature-request (https://github.com/AnySoftKeyboard/LanguagePack/issues/285) was going for.

nicoursi:
Fixes #2010
Fixes #1925
As the default qwertz and qwerty layouts didn't have any popups characters at all, I tried addressing this issue by just modifying those layouts instead of creating new ones.

If you think I should create new ones, leaving those original intact, I can do that.


addons/languages/german/pack/src/main/res/xml/de_qwerty.xml
```diff
@@ -1,48 +1,48 @@
...
-        <Key android:codes="119" android:keyLabel="w"  android:popupCharacters=""/>
-        <Key android:codes="101" android:keyLabel="e"  android:popupCharacters="€"/>
-        <Key android:codes="114" android:keyLabel="r"  android:popupCharacters=""/>
-        <Key android:codes="116" android:keyLabel="t"  android:popupCharacters=""/>
-        <Key android:codes="121" android:keyLabel="y" android:popupCharacters=""/>
-        <Key android:codes="117" android:keyLabel="u"  android:popupCharacters=""/>
-        <Key android:codes="105" android:keyLabel="i"  android:popupCharacters=""/>
-        <Key android:codes="111" android:keyLabel="o"  android:popupCharacters=""/>
-        <Key android:codes="112" android:keyLabel="p"  android:popupCharacters=""/>
+    <Row android:keyWidth="10%p">
```
menny: there are 11 keys in this row, it's should be set to 9.09%

menny: Unless, I'm missing something here?

addons/languages/german/pack/src/main/res/xml/popup_e.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
```
menny: I would change the name of this file to suggest it's for `de`. maybe `de_popup_e.xml`?

PR #2168, opened 2020-03-24
Auto capitalize autotext if typed word is capitalized

### LINKED ISSUE #2130: [Bug]english_autotext autocorrects in lower cases when the first letter needs to be uppercase
### Steps to reproduce
Type the following after a dot, question mark, or whatever situation requires the next letter to be in upper-case.

1. `Itll`
2. `Shouldnt`
3. `Teh`

### Expected behaviour
You should get the following words highlighted:
1. `It'll`
2. `Shouldn't`
3. `The`

### Actual behaviour
The highlighted suggestion will be in lowercases. This happens with every autotext words in [english_autotext.xml](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/addons/languages/english/pack/src/main/res/xml/english_autotext.xml#L96)
1. `it'll`
2. `shouldn't`
3. `the`

**Android OS version:**
Android 9

Xiaomi Redmi Note 7, lavender

English, Italian

nicoursi:
This fixes #2130 

It looks like it's working correctly now. The devs can have a look at this little piece of code, in case I missed some behaviour.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/dictionaries/Suggest.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/dictionaries/Suggest.java
@@ -342,8 +342,19 @@ public class Suggest {
                 mHaveCorrection = true;
                 if (mSuggestions.size() == 0) {
                     mSuggestions.add(mOriginalWord);
+                }           
+                //If the first character of the typed word is capitalized, 
+                //then also the suggestion will be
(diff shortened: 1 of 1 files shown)


menny:
@nicoursi code-format is invalid. You can automatically fix this by running `./gradlew googleJavaFormat`

PR #2174, opened 2020-03-25
put current layout language first in the dictionary overriding list

### LINKED ISSUE #1879: [Feature Request] override default dictionary - allow the user to decide what  dictionary should appear first on the list
### Steps to reproduce
1. Long press in enter,  then select at least two dictionaries from the list
2.write a new word in any app and  then double tap (on the suggestion)  to save it

### Expected behaviour
 There should be the possibility to choose which dictionary should be used to add new words in while typing. As, by default, new words are added to the first dictionary appearing in the list, the best/easiest way would be to give the user the possibility to sort that list in "override default dictionaries", so that he could decide what dictionary should appear first.

### Actual behaviour
The new word will be added to the language first in the list (in override default dictionary) even if the user wants differently (maybe adding it to the Italian dictionary and not the English one)

**Android OS version:**
9.0
Xiaomi Redmi Note 7
Italian, Czech, French, Spanish, English language packs

### LINKED ISSUE #2173: dictionary overriding problem [autototext related]
I would like to fix this issue myself but I need to be pointed at the right place.
### Steps to reproduce

1. Install `Italian` layout and select `English` and `Italian` in override default dictionary.
2. Type:
    `Mi piacciono i cani e i gatti`

### Expected behaviour
I should get:
    `Mi piacciono i cani e i gatti`

### Actual behaviour
I get: 
    `Mi piacciono I cani e I gatti`

This is so annoying that I have the dictionary overriding disabled, but I keep forgetting to switch language every time, so it is stil a nightmare

I'd like some advices on how I can address this issue. I'm not an advanced developer but I can find my way sometimes ;)

nicoursi:
Fixes #1879
Fixes #2173

<table>
<tr>
    <th>Enabled</th>
    <th>Disabled</th>
</tr>
<tr>
    <td></td>
    <td></td>
</tr>
</table>

This branch does two things:
1. Allows the user to add custom words to the language the current layout belongs to. The current layout  language always appears first  in the list now.
2. Adds a new checkbox in language settings, right after `Quick fixes and abbreviations`. It's unchecked by default. When checked, it disables autotext to every language checked in `Override default dictionary` except the primary one (current keyboard layout).
In this way, if you write in a Language that has lower-case `i` (ltalian and a few slavonic languages eg. czech) or other conflicting words/letters, you can disable autotext for secondary languages by  keeping it for the primary language (current keyboard layout).

Now I can go to italian layout and write:

1. `Mi piacciono i cani e i gatti and the sky is cloudy today.`

Then I switch to English layout and write:

1. `Ask is working all right now, do you think` **I** `should keep override default dictionary on?`

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
@@ -81,6 +81,9 @@ public class SuggestionsProvider {
     @NonNull private final List<EditableDictionary> mUserDictionary = new ArrayList<>();
     @NonNull private final List<NextWordSuggestions> mUserNextWordDictionary = new ArrayList<>();
     private boolean mQuickFixesEnabled;
+    // if true secondary languages will not have autotext on. For primary language is intended the
+    // current keyboard layout language
+    private boolean mQuickFixesSecondDisabled;
     @NonNull private final List<AutoText> mQuickFixesAutoText = new ArrayList<>();
 
     private boolean mNextWordEnabled;
@@ -131,6 +134,18 @@ public class SuggestionsProvider {
                                     mQuickFixesEnabled = value;
                                 },
                                 GenericOnError.onError("settings_key_quick_fix")));
+        mPrefsDisposables.add(
+                rxSharedPrefs
+                        .getBoolean(
+                                R.string.settings_key_quick_fix_second_disabled,
+                                R.bool.settings_default_key_quick_fix_second_disabled)
+                        .asObservable()
+                        .subscribe(
+                                value -> {
+                                    mCurrentSetupHashCode = 0;
+                                    mQuickFixesSecondDisabled = value;
+                                },
+                                GenericOnError.onError("settings_key_quick_fix_second_disable")));
         mPrefsDisposables.add(
                 rxSharedPrefs
                         .getBoolean(
@@ -264,7 +279,7 @@ public class SuggestionsProvider {
         mCurrentSetupHashCode = newSetupHashCode;
         final CompositeDisposable disposablesHolder = mDictionaryDisposables;
 
-        for (int i=0; i< dictionaryBuilders.size(); i++) {
+        for (int i = 0; i < dictionaryBuilders.size(); i++) {
             try {
                 Logger.d(
                         TAG,
@@ -281,7 +296,11 @@ public class SuggestionsProvider {
                 disposablesHolder.add(
                         DictionaryBackgroundLoader.loadDictionaryInBackground(cb, dictionary));
             } catch (Exception e) {
-                Logger.e(TAG, e, "Failed to create dictionary %s", dictionaryBuilders.get(i).getId());
+                Logger.e(
+                        TAG,
+                        e,
+                        "Failed to create dictionary %s",
+                        dictionaryBuilders.get(i).getId());
             }
 
             if (mUserDictionaryEnabled) {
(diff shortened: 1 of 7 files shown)


ime/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
```diff
@@ -264,60 +279,74 @@ public void setupSuggestionsForKeyboard(
         mCurrentSetupHashCode = newSetupHashCode;
         final CompositeDisposable disposablesHolder = mDictionaryDisposables;
 
-        for (DictionaryAddOnAndBuilder dictionaryBuilder : dictionaryBuilders) {
+        for (int i = 0; i < dictionaryBuilders.size(); i++) {
```
menny: I would add `DictionaryAddOnAndBuilder dictionaryBuilder = dictionaryBuilders.get(i);` and use `dictionaryBuilder` everywhere you have `dictionaryBuilders.get(i)`

ime/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
```diff
@@ -264,60 +279,74 @@ public void setupSuggestionsForKeyboard(
...
                 mUserNextWordDictionary.add(userDictionary.getUserNextWordGetter());
             } else {
                 Logger.d(TAG, " User does not want user dictionary, skipping...");
             }
-
-            if (mQuickFixesEnabled) {
-                final AutoText autoText = dictionaryBuilder.createAutoText();
+            // if mQuickFixesEnabled and mQuickFixesSecondDisabled are true
+            // it  activates autotext only to the current keyboard layout language
+            if ((mQuickFixesEnabled && mQuickFixesSecondDisabled && i == 0)
```
menny: this could be changed to:
```
if (mQuickFixesEnabled && (i == 0 || !mQuickFixesSecondDisabled)) {
```

ime/app/src/main/res/values/settings_keys_dont_translate.xml
```diff
@@ -354,6 +354,7 @@
     <string name="settings_key_persistent_layout_per_package_id_mapping">settings_key_persistent_layout_per_package_id_mapping</string>
 
     <string name="settings_key_quick_fix">quick_fix</string>
+    <string name="settings_key_quick_fix_second_disabled">quick_fix_second_disabled</string>
```
menny: settings keys should start with `settings_key_`. And in general, are the same string as the resource name. So: `<string name="settings_key_quick_fix_second_disabled">settings_key_quick_fix_second_disabled</string>`

ime/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
nicoursi: Thanks! This makes the code much simpler ;)

ime/app/src/main/java/com/anysoftkeyboard/dictionaries/SuggestionsProvider.java
nicoursi: I knew there should exist an easier way for that if line! Thank you

ime/app/src/main/res/values/settings_keys_dont_translate.xml
nicoursi: Done! I actually used `settings_key_quick_fix` as reference, and if you notice, its string is wrong as well. Should we change that, too?

nicoursi:
> Some comments. Also, this seems like something you want to have unit-tests for.
I have never done unit-test before. I could try having at look at it. Any links, advices, how-to are apprecited.

nicoursi:
@menny 
I wrote two unit tests. I don't know if they are good, as I'm very new to unit testing.

menny:
It's pretty good.

PR #2189, opened 2020-03-28
Added (Badly Implemented) Vietnamese Telex Typing

martholomew:
Phewf, this was/is a hard one.

I wrote some C to implement [Vietnamese telex typing](https://en.wikipedia.org/wiki/Telex_(input_method)), and connected the C to Java with the JNI.

The C is likely in a good location, but I put the Java in a not appropriate spot just for testing.

It still has a few problems:
* I'm not sure how to change the previous text without the suggestions disappearing, or how to change it at all.
* I'm not sure where the Java that implements the code should be.
* I'm not sure how to check which keyboard is being used so that telex is only enabled when it's Vietnamese.
* The C is still messy in places.
* The C makes Android throw a "StrictMode" error.
* I have no idea how to implement tests for this, lol.

@menny would you be able to help me out with a few things here?


menny:
If I understand correctly, what you are looking to implement is a character-sequence to a character.

Did you take a look at [this](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/java/com/anysoftkeyboard/keyboards/HardKeyboardSequenceHandler.java). It is used for physical keyboards, allowing you to map a sequence of key-presses to a character. Look for examples of a `PhysicalTranslation` xml files in language-packs.

It is possible to refactor this to work with on-screen keyboards, I'm sure.

martholomew:
Sadly, that won't work. Keys can change a letter 2 letters previous to the current one. So my solution, as of now, is to replace the current word with the edited one that the c outputs.

menny:
Can you share some examples?

martholomew:
[I have some tests that I run in my C](https://github.com/martholomew/telex-c/blob/master/telex.c#L212). "^", "/" and "~" among others represent combining diacritics since they are easier to see.

Word:  huye^n
Typed: r
Result: huye^>n

Word:  hon
Typed: s
Result: ho/n

Word:  huyen
Typed: x
Result: huy~en

Word:  huy~en
Typed: z
Result: huyen

Word: hu>o
Typed: n
Result: huo>n

martholomew:
Any thoughts on this? @menny

I'll likely be changing up the c a bit after looking at some other implementations.

menny:
is `telex` a Vietnamese thing only?

menny:
I feel like this is not related to anything other than Vietnamese. I wonder if we can think of a way to contain this logic somehow: either in the Vietnamese pack or as a "grammar-extension" inside the main-app that is enabled when Vietnamese is used.

martholomew:
Yeah, that would be optimal.

martholomew:
Is there any way currently to be able to put code of this kind inside of the language packs?

martholomew:
I'll bring this back when I've worked on it more.

PR #2195, opened 2020-03-31
Automatic spacing for punctuations added

nicoursi:
This feature takes care of spacing after and before punctuation. It extends the behaviour of `swap punctuation and space`

Here a few examples:

1. `"hi ,"`--> `"hi, "`
2. `"hi!"`--> `"hi! "`
3. `"hi !"` --> `"hi! "`
4. `"this is("` --> `"this is ("`
6. `"(good)"` --> `"(good) "`
7. `"(good )"` --> `"(good) "`

With French keyboard:
1. `"hi ,"`--> `"hi, "`
2. `"hi!"`--> `"hi ! "`
3. `"hi !"` --> `"hi ! "`
4. `"this is("` --> `"this is ("`
6. `"(good)"` --> `"(good) "`
7. `"(good )"` --> `"(good) "`

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
@@ -25,6 +25,7 @@ import com.anysoftkeyboard.dictionaries.DictionaryAddOnAndBuilder;
 import com.anysoftkeyboard.dictionaries.DictionaryBackgroundLoader;
 import com.anysoftkeyboard.dictionaries.Suggest;
 import com.anysoftkeyboard.dictionaries.TextEntryState;
+import com.anysoftkeyboard.dictionaries.TextEntryState.State;
 import com.anysoftkeyboard.dictionaries.WordComposer;
 import com.anysoftkeyboard.keyboards.AnyKeyboard;
 import com.anysoftkeyboard.keyboards.Keyboard;
@@ -145,6 +146,7 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
     private boolean mAllowSuggestionsRestart = true;
     private boolean mCurrentlyAllowSuggestionRestart = true;
     protected boolean mAdditionalCharacterForReverting;
+    private int mHowManyCharactersForReverting = 0;
 
     private boolean mJustAutoAddedWord = false;
     /*
@@ -661,6 +663,7 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
     }
 
     protected void handleSeparator(int primaryCode) {
+        mHowManyCharactersForReverting = 0;
         // Issue 146: Right to left languages require reversed parenthesis
         if (!getCurrentAlphabetKeyboard().isLeftToRightLanguage()) {
             if (primaryCode == (int) ')') {
@@ -699,6 +702,7 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
             // Picked the suggestion by a space/punctuation character: we will treat it
             // as "added an auto space".
             mAdditionalCharacterForReverting = !newLine;
+            if (!newLine) mHowManyCharactersForReverting++;
         } else if (separatorInsideWord) {
             // when putting a separator in the middle of a word, there is no
             // need to do correction, or keep knowledge
@@ -706,7 +710,7 @@ public abstract class AnySoftKeyboardSuggestions extends AnySoftKeyboardKeyboard
         }
 
         boolean handledOutputToInputConnection = false;
-
+        boolean sentTypedCharacter = false;
         if (ic != null) {
             if (isSpace) {
                 if (mIsDoubleSpaceChangesToPeriod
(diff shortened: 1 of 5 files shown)


nicoursi:
I will need to write some unit tests of course.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -86,6 +86,7 @@
 
     private boolean mAutoCap;
     private boolean mKeyboardAutoCap;
+    private boolean mShouldAutoSpaceForPunctuation;
```
menny, line 89: I believe it will make more sense to have this change done in the `AnySoftKeyboardSuggestions` class.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -662,6 +673,55 @@ private void onNonFunctionKey(
             default:
                 if (isWordSeparator(primaryCode)) {
                     handleSeparator(primaryCode);
+                    // retrieves the first two characters before cursor
+                    String previousTwoChars = ic.getTextBeforeCursor(2, 0).toString();
```
menny, line 677: accessing `ic` is a heavy operation (it basically, needs to communicate with the hosting app, and blocks the main thread).
Try to get that information either from `mLastSpaceTimeStamp`,`TextEntryState` or `mWord`.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -662,6 +673,55 @@ private void onNonFunctionKey(
...
+                    String stringPrimaryCode = new String(new int[] {primaryCode}, 0, 1);
+                    boolean anyPuntuaction =
```
menny, line 679: Can we use `com.anysoftkeyboard.ime.AnySoftKeyboardSuggestions#isWordSeparator` instead?

menny:
thanks!

If i think about this more, it seems that all of these should go under the usual `swapping punctuation and space`, and not as a new setting.

menny:
maybe try to expend `swapping punctuation and space` to support those cases?

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
nicoursi: I agree with you, that's why I started exactely from there but I got stuck. 

```
        if (TextEntryState.isPredicting() && !separatorInsideWord) {
            // ACTION does not invoke default picking. See
            // https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/198
            pickDefaultSuggestion(isAutoCorrect() && !newLine);
            // Picked the suggestion by a space/punctuation character: we will treat it
            // as "added an auto space".
            mAdditionalCharacterForReverting = !newLine;
```
Whatever modifications I make with `ic.deleteSurroundingText` and `ic.commitText` after `pickDefaultSuggestion` method call, will be applied before the `primarycode`. I didn't find any way  to add a space AFTER `primarycode` and not BEFORE. Whatever I do goes before!

Making changes in `AnySoftwareKeyboard` after `handleSeparator(primaryCode)` call magically works! :) I'm sure there is a way, please advice me :)

_If there in no other way_, I'd rather delete the `swapping punctuation and space` involved code and keep this new feature.

btw, I also extended the code to make it work with French punctuation if keyboard locale is fr ;) I'll push it after we decide how to apply the requested changes.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
nicoursi: as I commented above, as of now, I can only apply those changes after `handleSeparator` is been called. Before or inside `handleSeparator` didn't work for me. I need to undesrstand how to do that first.

nicoursi: sorry i didn't get your question right. 

`isWordSeparator` tells me if it is a separator or not. I need to know which separator is exactely. e.g `(` and `)` will be treated differently so I'm afraid isWordSeparator can't give me that information, as it is just a boolean.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
nicoursi: I think I got where I was in error. I have a better understanding of the code now. I can try to fix it now ;)

nicoursi:
@menny 
I tried to apply the changes you requested. This is not fully functional because I run into two problems.
The code basically does what is supposed to but:

1.  mLastSpaceTimeStamp is not reliable. There are cases where `mLastSpaceTimeStamp` should be `!= NEVER_TIME_STAMP` but it's not. 
This bug is already present in master, but probably nobody noticed. To see it just open ASK, and type `"word "`. Now type `"."` and you'll get `"word. "`. Now type again `"."` and you'll get `"word. ."` instead of `"word.." `. I tried to call `setSpaceTimeStamp(true);` every time I added a space with commitText, but it doesn't help.
If I use `ic.getTextBeforeCursor` to check for space, it all works as expected.

2. If I write `"hi "` and then `"!"` I'll get `"hi! "` as expected. Now, if i delete one character I'll get` "HHi"`. This happens everytime I delete one of the autospacing corrections. 
I made different attemps to troubleshoot this problem. I found out that, after autocorrection, I'm allowed to send only one character with  `sendKeyChar` or `ic.commitText`. If I send two (it's always the case with autospacing), then I get this problem. I really have no idea how to fix it!

3.  After this code will be fully functional and informally approved, I'll write the unit texts.

nicoursi:
I fixed everything now. It's all working. 

There's only one behaviour I don't know if it's ok:

- you type `"interesti"`, then you type `","` and you get `"interesting, "`.
So far so good! Now you try to delete the last character `" "` and you get  `"interesting,"` and not `"interesti"`. If this behaviour is not wanted, I don't really know how to fix it yet. See my comment above.

If you think this is good, let me know and I'll add/fix the unit texts

menny:
No, this should be fixed. Make sure the unit test passes.  I haven't taken a look at your code yet, my wife gave birth a few days ago and I'm not available for anything right now 😃

nicoursi:
> y wife gave birth a few days ago and I'm not available for anything right now smiley

Congratulations!!! No worries, enjoy the new entry of the family!

When you will have some time, feel free to give me some advices as I got a bit stuck there :stuck_out_tongue_winking_eye:

nicoursi:
This is fixed now. Everything works

nicoursi:
No interest from the maintainers, so I close this. Whoever is interested can fork it and improve it.

menny:
Can you rebase this? Sorry for the late response.

lubenard:
Hi, has this PR been integrated ? If not, i can have a look !

menny:
Please take a look.. I think you'll need to re-create this on your own branch

lubenard:
This is what i am doing, merging current master with his last commit.

He is using a lot of ```textEntryState```, which has been removed.

I am currently trying to find a solution for this.

menny:
that class was so terrible. I'm so happy it is removed.

PR #2207, opened 2020-04-06
English QWERTY with symbols on long press and numbers row

nicoursi:
This adds a new layout for English: `English with symbols and numbers row`.

This offers the chance to most symbols to be accessible as first choice on long press, and numbers will have their own key.

This is also for people who enjoy having a number row + a top row of their choice too.


nicoursi:
https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/ba6e1ed4c99b1f0ca85297707b2884a3472d1b12/ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardKeyboardSubtypeTest.java#L74-L83

Wow! I don't know why adding a simple English layout breaks so many tests!
If I'm understanding this right, I should add and `ID` into this array, but I don't really know where to get it from, as I don't know what a `Subtype` really is.

PR #2212, opened 2020-04-08
Generic rows 2

nicoursi:
This PR adds a new variant of the `Navigation keys alternative` top row with no numbers row on password mode. This is to make it compatible with the  the new coming PR: [2207](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2207)

It also removes emoji on long press on the space bar for the `Minimal` generic bottom row as this bottom row already has an emoji key.


nicoursi:
I forgot to leave a comment. This is working now,  unit tests pass with no problems.

PR #2222, opened 2020-04-13
coptic: Clean up play store listing

rillian:
Add screenshots and streamline the description.

Only an en-US localization available so far, but otherwise as far as I can tell coptic should now have a similar listing to other published addons.


addons/languages/coptic/apk/src/main/play/listings/en-US/full-description.txt
```diff
@@ -1,10 +1,9 @@
-Coptic Language pack
+Coptic Language input for AnySoftKeyboard
```
menny, line 1: Is there a benefit for using `input` rather than `pack`?

rillian: No. It seemed more descriptive, but I'm happy to use *pack* if you prefer.

There's a lot of variation among the various language addons. The most common seem to be `$langauge layouts and dictionary [for AnySoftKeyboard]` and `$language Language pack` Do you have a preference between those?

menny: Maybe `$language Language pack` is better, as it a more generic term.

menny: could you change that? And also rebase the branch

rillian: Ok, description changed and rebased against master.

menny:
thank you!

rillian:
Thanks for merging!

PR #2260, opened 2020-05-05
German Dvorak Compact Layout

xoomerq:
Hello,
I tweaked this nice English Dvorak-Compact layout into a German Layout.
Unfortunately, I am not a developer, so I apologise for my mistakes upfront. 
I am more a power user and tried my best at "copy&paste" :)

All what I did was:
-copy En Dv Compact to the german addons folder with an unique name,
-update german_keyboards.xml
-update lists with strings

Would it be possible to merge this, as it is, in the master branch?
Feedback appreciated!


menny:
Thanks! I'll let @martholomew review

PR #2282, opened 2020-05-21
QWERTY keyboard with only 26 letters

Roffild:
Without popups.

### DIFF
--- a/addons/languages/english/pack/src/main/res/values/english_pack_strings.xml
+++ b/addons/languages/english/pack/src/main/res/values/english_pack_strings.xml
@@ -1,6 +1,7 @@
 <resources>
     <string name="english_dictionary_description">English</string>
     <string name="english_keyboard_description">QWERTY Latin keyboard</string>
+    <string name="english_26letters_keyboard_description">QWERTY keyboard with only 26 letters</string>
     <string name="english_keyboard_symbols_description">QWERTY Latin keyboard with symbols (on long press)</string>
     <string name="dvorak_keyboard_description">Dvorak layout. Created with Stephen Paul Weber</string>
     <string name="compact_keyboard_description">Compact in portrait</string>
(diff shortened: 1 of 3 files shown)


menny:
Can you provide a screenshot of this?
Also, what is the purpose of such a keyboard?

Roffild:
You can click on the letter T, move your finger to the letter H and release. The letter H will be printed. On a standard keyboard, this will not make a popup on the letter G.

This keyboard is more convenient for the user who needs only 26 letters.

Additional characters are specified in the code and you can get rid of them only with: ``` android:popupCharacters="" ```
https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/8588455565b9709b24e5885143504de940a7b68d/ime/app/src/main/java/com/anysoftkeyboard/keyboards/ExternalAnyKeyboard.java#L426-L435

addons/languages/english/pack/src/main/res/values/english_pack_strings.xml
```diff
@@ -1,6 +1,7 @@
 <resources>
     <string name="english_dictionary_description">English</string>
     <string name="english_keyboard_description">QWERTY Latin keyboard</string>
+    <string name="english_26letters_keyboard_description">QWERTY keyboard with only 26 letters</string>
```
menny, line 4: can you change this to "QWERTY keyboard without any popups", or similar.

Roffild:
No matter what the description and name of this keyboard will be.

This UnitTest is impossible to pass:
```
com.anysoftkeyboard.ime.AnySoftKeyboardKeyboardSubtypeTest > testAvailableSubtypesReported FAILED
   java.lang.AssertionError: expected:<8> but was:<9>

com.anysoftkeyboard.ime.AnySoftKeyboardKeyboardSubtypeTest > testAvailableSubtypesReported FAILED
   java.lang.AssertionError: expected:<9> but was:<8>
```

nicoursi:
If you want to pass that awful test just have a look at how I did it here #2419

At one point, you will have to add and ID number into an array. Just invent one (no bigger then any number in that array) and add it as last number in the array. At this point you will fail the test, but you will know what ID to put (you'll find it in the error message). Make sure you put it at the right place, and delete the one you invented.

PR #2290, opened 2020-05-22
Nias language improved dictionary and new keyboard layouts

sslaia:
There are some important improvement in Nias language addon, mainly the dictionary and the keyboard layouts. I would be grateful if you can merge this. Best regards.


sslaia:
I could see that one check failed on buildSrc/src/main/java/github/RestRequestPerformer.java. I know I shouldn't touched that file as it exists outside the language addon I am working on.

However without the change in the file RestRequestPerformer.java (an added .toString() to StandardCharsets.UTF_8 in that file) Android Studio would refuse to sync gradle. Is there any solution for this?

sslaia:
Sorry I only found out just now that Java 10 is required for ASK. So reverted the commit and will start from scratch again. It means this pull request is redundant.

PR #2367, opened 2020-07-07
Add ask:isShiftAlways and Thai language keyboard

wishawa:
Add `ask:isShiftAlways` attribute to allow keyboard layout to specify whether a key's shift code should be used for sticky shift.
Add Thai keyboard (requires the `isShiftAlways` attribute to work properly).


addons/languages/thai/pack/src/main/res/xml/thai_keyboards.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Keyboards>
+    <Keyboard nameResId="@string/thai_keyboard" iconResId="@drawable/ic_status_thai"
```
menny, line 3: Can you attach two screenshots of these keyboards?

addons/languages/thai/pack/src/main/res/xml/thai_semialign.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:ask="http://schemas.android.com/apk/res-auto"
+    android:keyWidth="7.692307%p"
+    android:keyHeight="@integer/key_normal_height"
+    ask:showPreview="false"
+    ask:autoCap="false">
+    <Row>
+        <Key android:codes="3613" android:keyLabel="ฝ" ask:keyOutputTyping="ฝ" ask:isShiftAlways="true" ask:shiftedCodes="3622" ask:shiftedKeyLabel="ฦ" ask:shiftedKeyOutputTyping="ฦ" android:keyEdgeFlags="left" />
```
menny, line 10: What does this mean? in regards to `isShiftAlways=true`

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
```diff
@@ -830,14 +836,16 @@ public AnyKey(
...
 
-            // if the shift-character is a symbol, we only show it if the SHIFT is pressed,
-            // not if the shift is active.
-            mShiftCodesAlways =
-                    mShiftedCodes.length == 0
-                            || Character.isLetter(mShiftedCodes[0])
-                            || Character.getType(mShiftedCodes[0]) == Character.NON_SPACING_MARK
-                            || Character.getType(mShiftedCodes[0])
-                                    == Character.COMBINING_SPACING_MARK;
+            if (!mShiftCodesAlwaysOverride) {
```
menny, line 839: I think you don't need this extra field. You can either convert that into a variable in the method or change the condition to check `mShiftCodesAlways == false`

menny:
thank you for this contribution!!

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
wishawa: This way `isShiftAlways=false` can also be used to force the non-shift code when shift is only on sticky.

addons/languages/thai/pack/src/main/res/xml/thai_semialign.xml
wishawa: By ASK's default sticky shift does not apply to symbols or other non-letters. But Thai keyboards need sticky shift to apply to every key, so I introduce `isShiftAlways` to allow overriding the method ASK use to determine whether sticky shift should apply or not.

addons/languages/thai/pack/src/main/res/xml/thai_keyboards.xml
wishawa: Pics:

wishawa: I messed up the second one a bit. Pic is the fixed version, pushing the change now.

ime/app/src/main/java/com/anysoftkeyboard/keyboards/AnyKeyboard.java
menny: right, in that case, you can have `mShiftCodesAlwaysOverride` be a local method variable rather than a class field.

addons/languages/thai/pack/src/main/res/xml/thai_keyboards.xml
menny: thanks

wishawa:
@menny What need to be done to get the language pack app published?

menny:
I need to create the app in the play store and update all the required metadata. After that initial first publish it will be automatic .

PR #2372, opened 2020-07-09
Avoid "jacoco.exec: open failed: EROFS (Read-only file system)

ArenaL5:
The text in the commit is all I know. I'm not really all that familiar with Jacoco.

It's a problem I only found lately and seems to be caused by an upgrade in Jacoco. While I was typing, AnySoftKeyboard would randomly stop responding, but stay painted on the screen for 10 or 15 seconds (blocking the lower half of the screen), then close itself. Opening a logcat always showed the same error:
```
java.io.FileNotFoundException: /jacoco.exec: open failed: EROFS (Read-only file system)
```

(I neglected to save the rest of the logcat.)

I looked online for possible fixes to this problem and this solution seems to work, at least in my device. The file `jacoco.exec` is written in the same folder configuration files are exported to.

### DIFF
--- a/ime/app/src/main/resources/jacoco-agent.properties
+++ b/ime/app/src/main/resources/jacoco-agent.properties
@@ -1,0 +1,1 @@
+destfile=/sdcard/Android/data/com.menny.android.anysoftkeyboard/files/jacoco.exec


menny:
I'm assuming this happens only in debug builds?

ArenaL5:
Sorry I didn't answer earlier, I've been on a trip and forgot a bit about this.

I only use debug builds, so I'm not sure, but I think someone else would have noticed the error if it happened in other builds.

PR #2378, opened 2020-07-14
Fix adb helper script; tweaks to layouts; other commits ported from #1921

ArenaL5:
The primary goal for this commit is to fix `scripts/adb.sh`, a script that's really useful for me to debugging (and one I need to recreate every time I create a new branch). I'm probably the only person using that script, but it _does_ make things easier for me.

I also ported some micro-commits which I (wrongly) put in #1921, but are actually unrelated. I don't mind delaying the merge of that PR, it's a big change (and it might have a memory leak which I'm working on), but it also delays bug fixes for very minor, very long-lived, and annoying bugs. Most of them were marked as stale and closed, but aren't actually fixed.

My last commit is another self-serving addition, but I someone else will probably find it useful.

If you find it useful to merge this PR, I'll rebase #1921 to exclude all of these commits. #1921 will have only the custom ordering code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,14 @@ Sign up to beta-channel [here](https://play.google.com/apps/testing/com.menny.an
 
 ## Features
  * All kinds of keyboards:
-     * supporting lots of languages via external packages. E.g., English (QWERTY, Dvorak, AZERTY, Colemak, and Workman), Hebrew, Russian, Arabic, Lao, Bulgarian, Swiss, German, Swedish, Spanish, Catalan, Belarusian, Portuguese, Ukrainian and [many more](addons/languages/PACKS.md).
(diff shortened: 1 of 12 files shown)


menny:
what is that memory leak you mention?

menny:
I'm also going to remove the auto-stale-close mechanism

ArenaL5:
> what is that memory leak you mention?

Hours after fixing the jacoco error, I found that the keyboard kept crashing randomly. Last time, the logcat latest crash read something similar to this:

```
Process: com.menny.android.anysoftkeyboard, PID: 4123
    java.lang.OutOfMemoryError: Failed to allocate a 16 byte allocation with 8 free bytes and 8LB
```

It also said something about a target size I don't remember. The point is, it had never happened before, and I only remember making a variable non-static and some variables final.

I can't predict when it will crash. Sometimes it crashes mid-word, other times it crashes immediately when I open it after leaving the phone a couple hours in standby.

menny:
Does this crash happen on your branch only?

ArenaL5:
I couldn't verify that yet. I was hoping to merge this so it would be easier for me to test different branches.

I'm only changing one line of actual code (7f643d719fbf2d6610076f4bc186e13ead220df4), so this shouldn't make ASK any less stable.

menny:
thanks.
let me know what you find regarding that leak.

ArenaL5:
Will do. :relaxed:

PR #2406, opened 2020-07-29
Return Ctrl key back to inactive state after a key is pressed

mirfatif:


menny:
This PR can be closed, right? @mirfatif

mirfatif:
> This PR can be closed, right? @mirfatif

It needs to be merged. Without this CTRL key icon doesn't return back to inactive state after a key is pressed:

PR #2407, opened 2020-07-29
Make Ctrl work with non-terminal apps too

mirfatif:


menny:
this is great.
A couple of things:
1) can you add some comments and remove the extra logging?
2) can you add unit-tests for this?

Thank you

mirfatif:
@menny please check.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -667,15 +685,29 @@ private void onNonFunctionKey(
             default:
                 if (isWordSeparator(primaryCode)) {
                     handleSeparator(primaryCode);
-                } else {
-                    if (mControlKeyState.isActive() && primaryCode >= 32 && primaryCode < 127) {
+                } else if (mControlKeyState.isActive()) {
+                    int keyCode = getKeyCode(primaryCode);
+                    if (BuildConfig.DEBUG)
```
menny, line 690: No need for these logs. Please remove as many loggings as makes sense

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -667,15 +685,29 @@ private void onNonFunctionKey(
...
+                        Logger.d(TAG, "CONTROL state: Char code is " + primaryCode);
+                    if (Build.VERSION.SDK_INT >= 11 && keyCode != 0) {
+                        // TextView (and hence its subclasses) can handle ^A, ^Z, ^X, ^C and ^V
+                        // https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-10.0.0_r1/core/java/android/widget/TextView.java#11136
+                        int down = KeyEvent.ACTION_DOWN;
+                        int up = KeyEvent.ACTION_UP;
+                        int ctrlKey = KeyEvent.KEYCODE_CTRL_LEFT;
```
menny, line 697: Please use these constants directly

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -684,6 +716,8 @@ private void onNonFunctionKey(
                     } else {
                         handleCharacter(primaryCode, key, multiTapIndex, nearByKeyCodes);
                     }
+                } else {
+                    handleCharacter(primaryCode, key, multiTapIndex, nearByKeyCodes);
```
menny, line 720: Why do we need this call now? handleCharacter was not called before

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
```diff
@@ -1141,7 +1173,6 @@ public void onPress(int primaryCode) {
         if (primaryCode == KeyCodes.CTRL) {
             mControlKeyState.onPress();
             handleControl();
-            sendKeyDown(ic, 113); // KeyEvent.KEYCODE_CTRL_LEFT (API 11 and up)
```
menny, line 1144: So, we're holding up on this event until the user is pressing another key? What happens in the case where an app does want to react to CTRL press?

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
mirfatif: Done

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
mirfatif: Done

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
mirfatif: First call handles character when CTRL is pressed but character is out of range e.g. less than 32.
Second calls handles character when CTRL isn't pressed.

ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
mirfatif: I assumed that CTRL is always used with another key. That's how CTRL is used most of the times on Windows, Linux too. Reverted the changes back.

mirfatif:
> Also, do you think we need any additional unit-tests?

As I said, a mechanical engineer is not the right person to answer this. 🙂

Also you may make any changes you want (in this PR or others of mine) before merging. I've been using ASK for quite some while and made changes over time according to my needs. I pushed this thinking if my little contribution might add some value to this great project. You may drop the unwanted parts or change them as needed. I won't mind. I'm always available for assistance.

PR #2419, opened 2020-08-02
English qwerty with numbers row

nicoursi:
This adds a new layout for English: `English with symbols and numbers row`.

This offers the chance to most symbols to be accessible as first choice on long press, and numbers will have their own key.

This is also for people who enjoy having a number row + a top row of their choice too.


menny:
you know what I realize now... this could be useful for any layout. In other words: I think the `top-generic-row` should be a multiple selection thing, rather than a single-selection.
What do you say?

nicoursi:
>  think the top-generic-row should be a multiple selection thing, rather than a single-selection

I am not quite sure what you mean by that. If I understand that right, you mean that, from settings, you could add different top rows all together?

menny:
> > think the top-generic-row should be a multiple selection thing, rather than a single-selection
> 
> I am not quite sure what you mean by that. If I understand that right, you mean that, from settings, you could add different top rows all together?

exactly what I mean. Do you want to take a stab at that?

bugith:
Hi @menny & @nicoursi 
Sorry for intrusion in the conversation.
I'm currently working on azerty_accent.xml layout. I'm using the option to always show the numbers row on top of letters keyboard. I am not a developer so I very seldom use the symbols keyboard. In the need to use < & > for a comment in my PR, I notice that the first symbols keyboard shows two rows for numbers which is a bit... say... questionable:
Screenshot_20250919-112151



Please ignore if this is is unrelated.
Best regards

Issue #1682, opened 2019-03-29
Back-word gesture should not count spaces as a word


pieqq:
### Steps to reproduce
1. Go to `Settings > Gestures` and for `Swipe left`, select `Back-word`
2. Go to any app and start typing text, e.g. "This is a test"
3. Swipe the keyboard left **twice**

### Expected behaviour
Both "test" and "a" should be removed, and the remaining text should be "This is " (notice the final space, making it easier to immediately type new words)

### Actual behaviour
Only "test" and the space between "a" and "test" are removed. The remaining text is therefore "This is a".

### Rationale
Often, users want to remove several words at once. It's annoying to have to do the left swipe gesture for each spaces between words users want to delete.

I think back-word should remove any combination following the regexp `[\w"()'&,.?!-]+\s*` (that should be adapted to include all characeters with accents as well as additional typography such as French quotation marks `«»`, etc.

**Android OS version:** 8.0

**Device manufacturer and model:** Samsung Galaxy S7 Edge

**List of installed add-ons (like languages, or themes):** French keyboard

ArenaL5:
This issue was technically fixed; backword doesn't count spaces as a word now.

Currently it counts runs of letters (accented or not) and digits as words, but any punctuation breaks words (even apostrophes). I'd be happy to include some of the signs you suggested, but I'm not sure how useful would be all of them. Any other opinions?

pieqq:
Hello and thanks for the update! In what version of ASK was this fixed? I just tried again today with version 1.10.606 (from F-Droid) and it doesn't work.

> any punctuation breaks words (even apostrophes).

I think it's not optimal then. For instance, French uses a lot of apostrophes (e.g. "the tree" is « l'arbre » in French). English also uses apostrophes in the middle of what most people would consider a word, like "don't" or "it's". I believe in all these cases, most users would like a swipe left to remove the whole thing, not leave « l' », "don'" or "it'".

PR #2490, opened 2020-10-03
Make backword remove trailing spaces

maximbaz:
This implements part of #1682 that is about removing trailing spaces before actually removing a word.

The intended effect:

```
Before: "test this, " -> "test this," -> "test this" -> "test "
After:  "test this, " -> "test this" -> "test "
```

This actually makes the code respect the example that is already specified in the comment in the current code on line 928, this one:

> // 2b) "test this, " -> "test this"

What do you think?


menny (approved):

menny:
thank you!

PR #2520, opened 2020-11-05
Add SwiftKey-style bottom row

lucis-fluxum:
I added a SwiftKey-style bottom row with a slightly modified '.' key to replace a redundant comma with something cooler (the interrobang!)


ime/app/src/main/res/values/strings.xml
```diff
@@ -552,6 +552,7 @@
     <string name="extension_kbd_extension_numbers_symbols">Numbers and symbols</string>
     <string name="extension_kbd_bottom_minimal">Minimal</string>
     <string name="extension_kbd_bottom_minimal_no_emoji">Minimal (no emoji)</string>
+    <string name="extension_kbd_bottom_swiftkey">SwiftKey</string>
```
menny, line 555: I'm pretty sure I can not use this trademark in my repo. Can you think of a different description?

menny (changes requested):

ime/app/src/main/res/values/strings.xml
lucis-fluxum: Hmm, could we use "SwiftKey-style" or "SwiftKey-like"? Or something else entirely?

PR #2583, opened 2020-12-13
Numbers first in top row

Jonta:
I think the consistency of numbers being first is worth it

### DIFF
--- a/addons/languages/german/pack/src/main/res/xml/de_qwerty_slim.xml
+++ b/addons/languages/german/pack/src/main/res/xml/de_qwerty_slim.xml
@@ -6,15 +6,15 @@
     android:keyWidth="10%p">
     
     <Row android:keyWidth="10%p">
-        <Key android:codes="113" android:keyLabel="q" android:popupCharacters="\@1" android:keyEdgeFlags="left"/>
+        <Key android:codes="113" android:keyLabel="q" android:popupCharacters="1\@" android:keyEdgeFlags="left"/>
         <Key android:codes="119" android:keyLabel="w" android:popupCharacters="2"/>
-        <Key android:codes="101" android:keyLabel="e" android:popupCharacters="€3èéêëęē"/>
+        <Key android:codes="101" android:keyLabel="e" android:popupCharacters="3€èéêëęē"/>
         <Key android:codes="114" android:keyLabel="r" android:popupCharacters="4"/>
         <Key android:codes="116" android:keyLabel="t" android:popupCharacters="5"/>
         <Key android:codes="121" android:keyLabel="y" android:popupCharacters="6żžź"/>
-        <Key android:codes="117" android:keyLabel="u" android:popupCharacters="ü7ùúûũūŭű"/>
+        <Key android:codes="117" android:keyLabel="u" android:popupCharacters="7üùúûũūŭű"/>
         <Key android:codes="105" android:keyLabel="i" android:popupCharacters="8ìíîïłī"/>
-        <Key android:codes="111" android:keyLabel="o" android:popupCharacters="ö9òóôõōøœő"/>
+        <Key android:codes="111" android:keyLabel="o" android:popupCharacters="9öòóôõōøœő"/>
         <Key android:codes="112" android:keyLabel="p" android:keyEdgeFlags="right" android:popupCharacters="0"/>
     </Row>
     


menny:
@thomaseizinger @martholomew what do you think about this change? Germen users may find it strange that the order changed.

thomaseizinger:
> @thomaseizinger @martholomew what do you think about this change? Germen users may find it strange that the order changed.

Thank you for tagging me!

It has an IMO serious downside: how do I conveniently access `ü` and `ö` now?

I have an extra row of numbers on top, hence I don't quite understand why the number needs to come first on the long press.

The way I currently use the keyboard is by only having one and writing German and English with it. It is quite convenient not having to switch keyboards and still being able to access ä, ö, ü and ß.

If consistency is the reasoning we could also just remove the numbers from the top row :)

Also note that changing the long-click behaviour of `u` and `o` is inconsistent with `a` and `ß`.

Jonta:
> @thomaseizinger @martholomew what do you think about this change? Germen users may find it strange that the order changed.

(I came from a different keyboard, and found this strange. Not a big deal, got somewhat used to it after a few days, but worth noting)

> I have an extra row of numbers on top, hence I don't quite understand why the number needs to come first on the long press

I think the answer is in this sentence already: You're clarifying that you have an extra row of numbers on top, indicating that you know there are people who do not. =)

> It has an IMO serious downside: how do I conveniently access `ü` and `ö` now?
> 
> The way I currently use the keyboard is by only having one and writing German and English with it. It is quite convenient not having to switch keyboards and still being able to access ä, ö, ü and ß

I used to use (previous) keyboards like this: 3 languages on 1 layout, but then switched to using different ones because the number of plausible word-suggestions when swiping was too high with 3. 2 would make for a better experience of course

Iirc, I used the Norwegian layout (QWERTY with æøå-keys) at that time, and long-pressed aou to get äöü

In your case, I'd think a German layout with äöü-keys and a separate number row on top would work best, no? Or long-pressing aou, whichever you prefer

The consistency I advocate for does make significantly more sense to someone like me, who switches between layouts and therefore likes "long-press-U makes 7", than someone like you, who uses 1 layout all the time, and writes "ü" significantly more than "7". Especially because you have a whole additional key for 7

> If consistency is the reasoning we could also just remove the numbers from the top row :)

Then what of people who choose to not have a separate number row?

> Also note that changing the long-click behaviour of `u` and `o` is inconsistent with `a` and `s` and those can be argued should all be behave similarly

I'm not certain enough of what you mean to express agreement or disagreement, but generally like consistency in such cases

You two know better than I how ASK's different keyboard layouts interact with each other. From what @thomaseizinger writes, I get the impression that "German" and "German w/number row on top" is an extension of German. If so: Perhaps it makes more sense to decouple the two

And do we at least agree that 1 and 3 should change? :)

thomaseizinger:
> > I have an extra row of numbers on top, hence I don't quite understand why the number needs to come first on the long press
> 
> I think the answer is in this sentence already: You're clarifying that you have an extra row of numbers on top, indicating that you know there are people who do not. =)

What I meant to express with this is that there is already a setting in ASK (unconditionally from the layout) to add a "generic row" at the top with just numbers. If you want easy access to numbers, that would IMO be the "ASK-way" to do it.

> > If consistency is the reasoning we could also just remove the numbers from the top row :)
> 
> Then what of people who choose to not have a separate number row?

Alternatively to the additional row, you can also switch to the symbols page to get access to the numbers among a whole lot of other symbols.

> In your case, I'd think a German layout with äöü-keys and a separate number row on top would work best, no? Or long-pressing aou, whichever you prefer

I prefer long-pressing instead of additional keys because it mimics the layout I have on my laptop and desktop computer :)
In fact, the reason I use the German QWERTY layout is because long-pressing gives me easy access to those additional keys.

> > Also note that changing the long-click behaviour of `u` and `o` is inconsistent with `a` and `s` and those can be argued should all be behave similarly
> 
> I'm not certain enough of what you mean to express agreement or disagreement, but generally like consistency in such cases

I meant to express that currently, the keyboard is consistent in that long-clicking a,s,u,o gives you the additional letters of the German alphabet ä,ß,ü,ö. That consistency is removed if we introduce consistency in regards to the numbers row. As such, it is difficult to say that making this change makes things more consistent :)

> And do we at least agree that 1 and 3 should change? :)

I am not fussed about 1 and 3 but to be honest, but thinking about it now, it seems removing the numbers would actually make things "better" because it removes the inconsistency in how to access them. As mentioned above, ASK already provides a way for easy access to numbers: a "generic top row". The setting is a bit hidden but it exists!

Jonta:
**tl;dr at bottom**

> What I meant to express with this is that there is already a setting in ASK (unconditionally from the layout) to add a "generic row" at the top with just numbers. If you want easy access to numbers, that would IMO be the "ASK-way" to do it.

*The* ASK-way, or *an* ASK-way?

I assume there's some reasoning behind providing number-keys on long press of top row in at least some cases. :)

> Alternatively to the additional row, you can also switch to the symbols page to get access to the numbers among a whole lot of other symbols.

Sure. But who wants to do that when typing <5 digits?

> I prefer long-pressing instead of additional keys because it mimics the layout I have on my laptop and desktop computer :)
> In fact, the reason I use the German QWERTY layout is because long-pressing gives me easy access to those additional keys.

I don't understand. Pretty much all pictured here have separate keys for äöü: https://en.m.wikipedia.org/wiki/German_keyboard_layout

(FWIW: On PC, I prefer pressing a key before "aou" to add an umlaut ( ¨ ) on top. Part of the Norwegian layout, which is the one I always use on PC)

> I meant to express that currently, the keyboard is consistent in that long-clicking a,s,u,o gives you the additional letters of the German alphabet ä,ß,ü,ö. That consistency is removed if we introduce consistency in regards to the numbers row. As such, it is difficult to say that making this change makes things more consistent :)

I want more consistency between different layouts
You want more consistency with… hm. What you're used to (I don't want to put words in your mouth)

> I am not fussed about 1 and 3 but to be honest, but thinking about it now, it seems removing the numbers would actually make things "better" because it removes the inconsistency in how to access them. As mentioned above, ASK already provides a way for easy access to numbers: a "generic top row". The setting is a bit hidden but it exists!

That is true
But we also know that very few people change their settings. A biggger percentage when they're someone who downloads a different keyboard? For sure. But ASK might the default in places

**Big picture:**

I think the best solution here would be a separate German keyboard layout, that is more consistent with other layouts. Number row as part of the QWERTY-row, with numbers as priority

What's ASK's threshold for adding a new one? Rules, guidelines? :)

thomaseizinger:
> > I prefer long-pressing instead of additional keys because it mimics the layout I have on my laptop and desktop computer :)
> > In fact, the reason I use the German QWERTY layout is because long-pressing gives me easy access to those additional keys.
> 
> I don't understand. Pretty much all pictured here have separate keys for äöü: [en.m.wikipedia.org/wiki/German_keyboard_layout](https://en.m.wikipedia.org/wiki/German_keyboard_layout)

I use an English keyboard everywhere. The German layout is horrible for programming :)

> I assume there's some reasoning behind providing number-keys on long press of top row in at least some cases. :)

I guess so! If you want to save vertical space, then having a dedicated row with numbers all the time can be annoying!

> > Alternatively to the additional row, you can also switch to the symbols page to get access to the numbers among a whole lot of other symbols.
> 
> Sure. But who wants to do that when typing <5 digits?

Personally, I switch to the symbols page all the time for typing single symbols (`!` for example). ASK doesn't show long-press options until you actual press it. I believe the Google keyboard shows the long-press symbols in a tiny font so you don't have to guess which one it is. Apart from the obvious ones that we are discussing here (`ä`, `ö`, `ü`, `ß`) it seems arbitrary which symbol hides under which letter when you long press. As such, I don't even try to remember it and use the symbols page right away :sweat_smile: 

> > I meant to express that currently, the keyboard is consistent in that long-clicking a,s,u,o gives you the additional letters of the German alphabet ä,ß,ü,ö. That consistency is removed if we introduce consistency in regards to the numbers row. As such, it is difficult to say that making this change makes things more consistent :)
> 
> I want more consistency between different layouts
> You want more consistency with… hm. What you're used to (I don't want to put words in your mouth)

I want consistency _within_ the layout so that there is a predictable way of typing the aforementioned letters.

> I think the best solution here would be a separate German keyboard layout, that is more consistent with other layouts. Number row as part of the QWERTY-row, with numbers as priority
> 
> What's ASK's threshold for adding a new one? Rules, guidelines? :)

In my experience, maintainers are usually happy to merge new layouts, however my question would be: What would be "German" about this layout if it no longer favors `ä`, `ö`, `ü`, `ß` over other symbols?

QWERTY by itself is already not standard German. If the four additional letters of the German alphabet are not favored, what is the difference to - for example - the standard English Latin layout?

I just checked and that one allows you to type `ä`, `ö`, `ü`, `ß` although they are not pleasant to access (because they are not the standard long-press option).

To me, the German QWERTY layout provides value because:

- QWERTY is what I use across all my devices
- I can easily access the four additional letters without having to "search" for them in the long-press popup

Esp. the latter is IMO the primary value-add of the German keyboard over the standard English one.

As such, I agree that a new layout might be the best choice here.

menny:
this discussion reminds me a change suggested by @ArenaL5 where the user could decide where the digits are positioned in the long-press.
@ArenaL5 am I right?

ArenaL5:
Yes, you're speaking of PR #1921. Glad you mentioned me, I have some news.

I've left that PR on hold because I suspected it was causing a memory leak (and also because I was very busy with my offline life).

I haven't tested that lately, but I kept using AnySoftKeyboard, and right now my main suspects are the gesture dictionaries. For a long time, I couldn't type anything while the gestures were loading, or the keyboard would freeze and then crash. (I didn't file an issue yet because I can't find a reliable way to replicate it anymore.)

PR #2586, opened 2020-12-13
Numbers first in top row

Jonta:
I think the consistency in numbers being first is worth it

### DIFF
--- a/addons/languages/german/pack/src/main/res/xml/de_qwerty_extra.xml
+++ b/addons/languages/german/pack/src/main/res/xml/de_qwerty_extra.xml
@@ -6,15 +6,15 @@
     android:keyWidth="10%p">
     
     <Row android:keyWidth="10%p">
-        <Key android:codes="113" android:keyLabel="q" android:popupCharacters="\@1" android:keyEdgeFlags="left"/>
+        <Key android:codes="113" android:keyLabel="q" android:popupCharacters="1\@" android:keyEdgeFlags="left"/>
         <Key android:codes="119" android:keyLabel="w" android:popupCharacters="2"/>
-        <Key android:codes="101" android:keyLabel="e" android:popupCharacters="€3èéêëęē"/>
+        <Key android:codes="101" android:keyLabel="e" android:popupCharacters="3€èéêëęē"/>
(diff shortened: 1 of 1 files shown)


menny (commented):
These exceptions were requested by users, as those letters are super popular and it makes sense to keep them before the digits.
What do you think?

Jonta:
> These exceptions were requested by users, as those letters are super popular and it makes sense to keep them before the digits.
What do you think?

@menny Was that directed at me?

Nobody else has responded so far, and according to GitHub's webinterface, we're the only 2 participants in https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2586

What changes are you referring to?

To me, it looks like you're asking me whether I agree with having some ("those") letters before digits, as they're popular

This confuses me. Especially with my consistency argument

I did have a conversation with someone about this here on GitHub, around the time I submitted this pull request

menny:
I think it's best to get some feedback from a large set of users around this.
Maybe we can find reviews in the Play Store. What do you say?

Jonta:
I'm thinking there's a table of yes/no-switches people can have with this:
- ä, ö, ü each have their own, dedicated button
- An Umlaut-button to modify the next typed a, o, u
- Have ä, ö, ü pop up when long-pressing a, o, u
- ä, ö, ü come up when long-pressing top-row
- ä, ö, ü come up before numbers when long-pressing top-row

That's already 2^5 = 32 combinations

With my setup, to get top row numbers in German QWERTY (ä, ö, ü have their own dedicated buttons. Number row not visible), I have to do this after long-pressing:

(6: Number-layout)

| Letter | Action  |  Result |
|---|---------|---------|
| Q | 1 right |       1 |
| W | Let go  |       2 |
| E | 1 up    |       3 |
| R | Let go  |       4 |
| T | Let go  |       5 |
| Y | N/A     | Nothing |
| U | 1 left  |       7 |
| I | Let go  |       8 |
| O | 1 left  |       9 |
| P | Let go  |       0 |

And that's rather annoying. =)

I'm probably using a different German-layout, seeing as 6 _is_ supported in this one's number row

Jonta:
Ideally, all 32 (or more) combinations should be supported

If we can get information about which people like best, that would be great. Fantastic default settings FTW

PR #2590, opened 2020-12-14
New Language: Portuguese

somini:
This is `pt-PT`. `pt-BR` is already available, but this is different enough to warrant a new variant.

I mostly migrated the abandoned LanguagePack, and cleaned up some images and small tweaks. I have some other sources for better dictionaries, but it's 4 AM now, I can add this later.

I also added a new layout that includes symbols, to match the features from other languages.

I followed the CONTRIBUTING file, thanks for that documentation. Sounded like a daunting task, but it's actually simple, very well organized. The hardest part is to get the Android stuff without downloading 7GB.

This is tested on a crappy old phone I had lying around, added screenshots for that. Feel free to publish this to F-Droid and Play Store under the ASK organization.

Only a draft because I'm not sure I missed some things, CI will tell.


menny:
I see that you used a different  application ID (`portuguese2` instead of `portuguese`). Any reason why you do not want to replace https://play.google.com/store/apps/details?id=com.anysoftkeyboard.languagepack.portuguese ?

somini:
> I see that you used a different application ID (`portuguese2` instead of `portuguese`). Any reason why you do not want to replace https://play.google.com/store/apps/details?id=com.anysoftkeyboard.languagepack.portuguese ?

Not reason, since it already existed built from another source, I though I couldn't really "overwrite" it. I saw "russian2" and though this was a similar case.

I can change the IDs to the original version,  just let me know.

somini:
Renamed the ID.

menny (approved):

menny:
thank you!

telmobss1:
The native br layout is not yet available in Br- is an adaptation of the keyboard to the Brazilian language

telmobss1:
https://pt.m.wikipedia.org/wiki/BR-Nativo

PR #2597, opened 2020-12-16
Squash several minor bugs

ArenaL5:
This closes several minor issues, some of them very old. If nothing else, this should remove a couple quirks of the keyboard and shorten a little bit the issue list. (And also, add a new keyboard layout for Danish, thanks to @snkfxx.)

The greatest change is that “Reverse cycle through all keyboard” and “Cycle through all keyboards” are finally different actions, so I needed to add extra code.

Every commit is tagged with the issue it fixes.


menny (approved):

menny:
these are great!
Thank you!

PR #2612, opened 2020-12-31
add KOY variant of AdNW to German language pack

duchampdev:
This PR contains the KOY-variant of the AdNW layout (w/ some minor changes, see below) in the German language pack.
For ease of use, I used an ortholinear layout like some ergonomic keyboards do and rearranged one key. Furthermore, I moved the shift keys to the bottom row, sacrificing the up/down arrows. This way, broader keys (9% vs. 8%) are possible. I found this alternative to be a rather minor change compared to the normal key alignment of AdNW in ASK but being much more usable.
If you wish, I can also add a layout with the default arrangement. And in general, just let me know, if you'd like to see something changed in order to make the PR more adequate to merge, @menny .

### DIFF
--- a/addons/languages/german/pack/src/main/res/xml/de_adnw_koy.xml
+++ b/addons/languages/german/pack/src/main/res/xml/de_adnw_koy.xml
@@ -1,24 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- Copyright 2016, kertase, 2020, chrbauer, 2020 duchampdev
 
-    This file is free software: you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation, either version 3 of the License, or
-    (at your option) any later version.
-
-    This program is distributed in the hope that it will be useful,
-    but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-    GNU General Public License for more details.
-
-    You should have received a copy of the GNU General Public License
-    along with this program.  If not, see <http://www.gnu.org/licenses/>. -->
 <Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:ask="http://schemas.android.com/apk/res-auto"
           android:keyWidth="9%p" android:keyHeight="50dip"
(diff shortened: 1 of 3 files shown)


addons/languages/german/pack/src/main/res/xml/de_adnw_koy.xml
```diff
@@ -0,0 +1,85 @@
...
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
```
menny, line 14: I'm not sure what does including a GPL licensed file in an Apache2 codebase means. Do you?

menny:
Thank you for this layout. I would like to merge but not sure about the license and the copyright.

duchampdev:
@menny Thanks for your kind feedback! I used `de_adnw.xml` as the basis for `de_adnw_koy.xml`, as there are many similarities between both layouts. Therefore, I also copied over the license header present in this file and added my GitHub name there because I thought it would be intended like that. The GPL header is present both in the adnw and the neo layout file(s).

However, I just looked at the qwertz layout file and saw that it does not contain a license header. So I would like to remove it from my file as well, because as far as I understand [this](https://www.apache.org/licenses/GPL-compatibility.html), GPLv3 inside an Apache codebase is problematic. However, as I used `de_adnw.xml` to create `de_adnw_koy.xml`, I'm afraid just removing the header violates the license of `de_adnw.xml`.

addons/languages/german/pack/src/main/res/xml/de_adnw_koy.xml
menny: please remove the header. As per https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/CONTRIBUTING.md#license

duchampdev: Done!

menny (approved):

PR #2622, opened 2021-01-09
Nias keyboard layout improvements

sslaia:
Minor layout improvement


sslaia:
Sorry for all there errors and I don't know how to solve them. I only updated the keyboard layout (xml file) in the nias addon language and the errors are about issue like "Project 'ime' not found in root project 'workspace'" which I haven't touched at all.

Should I close/withdraw this pull request?

addons/languages/nias/pack/src/main/res/xml/nias_keyboards.xml
```diff
@@ -9,23 +9,5 @@
         index="1"
         layoutResId="@xml/nias_qwerty"
         nameResId="@string/nias_qwerty" />
-    <Keyboard
```
menny, line 12: Why are you removing these layouts?

settings.gradle
```diff
@@ -1,75 +0,0 @@
-rootProject.name = 'AnySoftKeyboard'
```
menny, line 1: Why are you removing this file? It is very important :)

menny (changes requested):

addons/languages/nias/pack/src/main/res/xml/nias_keyboards.xml
sslaia: Those extra layouts were not used and have confused users. The standard one is now improved based on feedback I received. One keyboard layout with characters ö and ŵ is enough for Nias language.

settings.gradle
sslaia: I really don't know how that happened. I check my seetings.gradle and the line is there. I only commented out other addon languages so that they are not compile every time I create a build for testing.

I am really sorry for this and I don't know how to resolve it as in my settings.gradle that line is still there. Should I just cancel/close the pull request?

sslaia: Let me close this request first. I noticed that Android Studio has different git status (with old deleted branches) than in command line, where I need to build because of the Java issue in Android Studio. I need to sort this out first before pull another request.

PR #2626, opened 2021-01-10
Set to capitalize when deleting whole line with shift + backspace

lubenard:
When using the Shift + backspace shortcut to delete the whole line, the shift state was not resetting to enabled, meaning you had to reclick on the shift key. 

This PR take care of this.

I do not know if my code is the most optimised way to do it. I am maybe missing something, and i do not like calls to ```ic.getTextBeforeCursor```

Maybe there is a another way to do it ?

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java
@@ -91,6 +91,8 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardColorizeNavBar {
 
     private int mOrientation = Configuration.ORIENTATION_PORTRAIT;
 
+    private int mShiftBackspaceLenghtRemaining = -1;
+
     private static boolean isBackWordDeleteCodePoint(int c) {
         return Character.isLetterOrDigit(c);
     }
@@ -893,6 +895,11 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardColorizeNavBar {
             currentComposedWord.reset();
             getSuggest().resetNextWordSentence();
             ic.setComposingText(textAfterCursor, 0);
+
+            // I do not know if we can avoid call ic because it is a heavy call.
+            // TextAfterCursor is only the last typed word
+            mShiftBackspaceLenghtRemaining = ic.getTextBeforeCursor(128, 0).length();
+
             postUpdateSuggestions();
             return;
         }
@@ -901,6 +908,8 @@ public abstract class AnySoftKeyboard extends AnySoftKeyboardColorizeNavBar {
         // Which is always good. This is a part of issue 951.
         CharSequence cs = ic.getTextBeforeCursor(128, 0);
         if (TextUtils.isEmpty(cs)) {
+            // We handle the case where there is nothing to delete anymore
+            mShiftBackspaceLenghtRemaining = 0;
             return; // nothing to delete
         }
         // TWO OPTIONS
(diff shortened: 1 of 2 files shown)


menny:
I can suggest a different approach that does not require `getTextBeforeCursor` and does not require to keep a count. I think it will work, not sure:
When SHIFT+DELETE is pressed (specifically, at `com/anysoftkeyboard/AnySoftKeyboard.java:394`), you should flip a flag, call it `mShouldUpdateCapsAfterShift`.
Once the shift key is released (maybe at `onRelease(primaryCode)`?), check for that flag and call `updateShiftStateNow()`.

Now, I see that there is some logic around calling `updateShiftStateNow()` in `onRelease`, but I'm not sure what that suppose to do, and if we can re-use (or extend) that logic to also support your case

menny:
You might also need to add support for `ic.getCursorCapsMode` in `TestInputConnection`, because right now it always returns `0`. See com.anysoftkeyboard.TestInputConnection#getCursorCapsMode

lubenard:
Hello ! After futher investigations, it seems that modifying getCursorCapsMode will break some tests (About 100).

This is due to the fact that using a "Normal" keyboard, saying ```mAnySoftKeyboardUnderTest.simulateTextTyping("hello");```

would result into ```Hello``` and not ```hello```.

Currently, and after many documentation ([here](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/inputmethod/BaseInputConnection.java;bpv=1;bpt=1;l=469?gsn=getCursorCapsMode&gs=kythe%3A%2F%2Fandroid.googlesource.com%2Fplatform%2Fsuperproject%3Flang%3Djava%3Fpath%3Dandroid.view.inputmethod.BaseInputConnection%23586999e4b809d03a89a0a6ee367031307712ee43e857556545184251b9909884)), it seems ```ic.getCursorCapsMode``` is calling ```TextUtils.getCapsMode```. 

My modified version of getCursorCapsMode is currently looking like this:

```
@Override
    public int getCursorCapsMode(int reqModes) {
        // Number 147457 come from real test using real device.
        // Seems mocked number set reqModes to '1', leading it to fails every time.
        return TextUtils.getCapsMode(mInputText.toString(), mCursorPosition, 147457);
    }
```

But, it is working in real conditions ! No more ```ic.getTextBeforeCursor``` ! 
What do you think ?
Should we modify all of those tests ?

It seems to be the cleanest version

menny:
so, how does your change looks like now? Can you push it here?

lubenard:
Hey ! I was waiting for your answer for this
 ```
What do you think ?
Should we modify all of those tests ?
```
, but i can try to push something tonight !

**Comment by menny** [comment]
Yes, please change the implementation of `TestInputConnection` to use `TextUtils.getCapsMode`.
And fix all the affected tests.

I would also love it if you add a test that ensures that the settings `R.string.settings_key_auto_capitalization` affects this as well.

**Comment by lubenard** [comment]
Hey ! I changed all tests but some things triggers me.
For example, suggestions are lowercase even if the word is having a uppercase letter

For example, suggestions for "He" are  :    "He", "he'll", "hell", "hello" 

and i can't figure out why the other suggestions have a H in lowercase

Second thing: 

When the test use simulateGestureProcess, the next characters are Uppercase.

Example,

```
simulateGestureProcess("hello");
Assert.assertequals("Hello", mAnySoftKeyBoardUnderTest.getCurrentInputConnectionTest()); <----- True

mAnySoftKeyBoardUnderTest.simulateTextTyping("a");
Assert.assertequals("Hello a", mAnySoftKeyBoardUnderTest.getCurrentInputConnectionTest()); <----- False, "Hello A" is actual text.

```

This seems to act like this only for simulateGestureProcess.

I thought it might be related to policy for getCapsMode ( TextUtils.CAP_MODE_CHARACTERS, TextUtils.CAP_MODE_WORDS, or TextUtils.CAP_MODE_SENTENCES)

But that does not seems related

PR #2652, opened 2021-01-25
Added an English QWERTZ keyboard

Apflkuacha:
Added an English QWERTZ keyboard with symbols as popups, very similar to the Gboard layout. In the German comments on PlayStore some were complaining that there is no WERTZ keyboard for English language. 
By the way: do you need someone who responds to these comments in German? Some people complain about stuff which could be changed in the settings. They couldn't find it or haven't even had a look the settings..


menny (approved):
thank you

menny:
I would love to get some help with replies and support. How do you suggest we do that?

Issue #2630, opened 2021-01-15
[Feature Request] Add Option to Disable Key Popup for Password Fields


hbchai:
The keyboard currently pops up the pressed key on all fields, including password fields. This makes it easier for an onlooker to steal a password.

This request is to add an option to avoid popping up the pressed key on password fields.

menny:
this is a good suggestion. I'll try to get this done for r4, but may need to push it to r5.

PR #2654, opened 2021-01-26
No popups with passwords

menny:
fixes https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/2630


menny:
While coding this, I found a bug with the popup reuse algorithm, so a double win here!

Issue #2623, opened 2021-01-09
Extra space when using autosuggestion and quotation marks


thomascrisan:
### Steps to reproduce
1. Begin typing a word in quotation marks.
2. Pick an autosuggested word.
3. Type the closing quotation mark.

### Expected behaviour
Remove the space after autosuggested word so that the word is properly enclosed by the marks, for example: "quotation“

### Actual behaviour
It looks like this: "quotation “

**Android OS version:**
7.01

**Device manufacturer and model:**
Xiaomi Mi Max

**List of installed add-ons (like languages, or themes):**

joe540:
You can disable that behavior by going into the settings, Language > Tweaks and more > Tweaks > Uncheck "Space after candidate pick"

Hope that helps!

thomascrisan:
Having a space after the candidate pick is the desired behavior except after when using some types of punctuation such as a period ending a sentence.  After selecting a candidate word and then typing a period, the ending space is swapped with period if the option "Swap punctuation and space" is checked.  I am asking that the same action be performed with a quotation mark.

joe540:
Sorry, I misunderstood your first message. I proposed a PR that would include ' and " as punctuations.

HubbyG:
1) swipe word, select non highlighted suggestion. Swipe another - extra space inbetween!
2) open ( bracket. Swipe word. Close bracket). Extra space after'(' ( but not before) ')'

decadent:
Btw, recognized quotes should properly depend on the current language—i.e. perhaps it's best to have plugins list the quote characters. Because some languages have the characters reversed, e.g. “English” vs „Russian“, «French» vs »German«. That is, if ASK is going to distinguish opening and closing quotes, which would be nice for best results.

In some languages, opening and closing quotes are identical—like ”Finnish”—which I guess reverts to the case of "straight quotes".

joe540:
Language pluguins can already overwrite the DEFAULT_SENTENCE_SEPERATOR

Note that as a french speaker, I never use « and » but use ". We can add special punctuations per language but should keep some basic ones like ' and " in the general punctuations

decadent:
> There is a naive, and expensive, solution that when a user `"` after an auto-space AnySoftKeyboard will scan the previous characters for an openning `"`

Is it really expensive, though? I would assume that modern-ish phones would have no problem iterating through, say, a hundred characters. Especially since user input is generally slow compared to computing.

[...]

PR #2705, opened 2021-02-15
Add single and double quotes to DEFAULT_SENTENCE_SEPARATORS

joe540:
Proposal for issue #2623 by adding single quotes and doubles quotes to DEFAULT_SENTENCE_SEPARATORS

Note that for sentenceSeparators to be updated, the application data needs to be wiped (if the app settings are restored from a backup, sentenceSeparators will be restored also)

I think quotes (both singles and doubles) pretty much separate a sentence the same way a parenthesis does, so I think it would make sens to add them to DEFAULT_SENTENCE_SEPARATORS.

Otherwise, another way to fix it would be to add a seperate check in isSpaceSwapCharacter().

That's my first PR so I may not be doing it correctly
Let me know if you want me to change anything

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardFactory.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardFactory.java
@@ -36,7 +36,7 @@ public class KeyboardFactory extends AddOnsFactory.MultipleAddOnsFactory<Keyboar
     private static final String XML_ADDITIONAL_IS_LETTER_EXCEPTIONS_ATTRIBUTE =
             "additionalIsLetterExceptions";
     private static final String XML_SENTENCE_SEPARATOR_CHARACTERS_ATTRIBUTE = "sentenceSeparators";
-    private static final String DEFAULT_SENTENCE_SEPARATORS = ".,!?)]:;";
+    private static final String DEFAULT_SENTENCE_SEPARATORS = ".,!?)]:;'\"";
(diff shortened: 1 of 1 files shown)


menny:
Thanks for this PR. 
1. Why do you say that "the application data needs to be wiped"?
2. Can you submit this PR to the release branch? Instead of `master` target the PR to `release-branch-ime-v1.10-r4`

joe540:
>     1. Why do you say that "the application data needs to be wiped"?

From my tests, "sentenceSeparators" seems to be saved as a setting somehow. I needed to remove the app data for that line to be run:

https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/dcf82edd080e56c319a438b64cd3f060132ee745/ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardFactory.java#L86

Once I restored my settings backup, it seems to have restored my old sentenceSeparator (without ' or ") even if I didn't find any reference in the xml file..

joe540:
OK, I just realized that this will not work properly. As ' and " are used both to start and end a quotation, we can't assume that ' or " would be meant to be at the end of the previous word, rather than at the beginning of the next one.

If I want to type `I am 'joe'`, for example, and that I select "am" from the suggestions, it would write ` I am' joe'`. It's also true for some other cases like using the smiley ): after selecting a suggestion which would also result in  `I am sad) :`  instead of `I am sad ):`, but it's way less likely to happen than with ' and "

menny:
that's true.. too bad, could have been a good fix.

Thanks.

Issue #2252, opened 2020-04-29
[feature request] ordinal symbols in PT-PT


1000283:
Physical pt-pt keyboards and most virtual keyboards have male and female [ordinal symbols](https://en.wikipedia.org/wiki/Ordinal_indicator).
They have a dedicated key on physical keyboards or are usually on the alternative characters for A and O letters.
It would be neat if these were added to pt-pt as they are quite common.

maverick74:
Indeed they are very used!!!

PR #2730, opened 2021-02-27
fix to #2252 added fe/male ordinals to pt keyboard

1000283:
The ordinals ("first" can be written as "1º" or "1ª" depending on gender) are in very common use in Portuguese, even having a dedicated key on Portuguese keyboards.
In software keyboards they are usually implemented as alternative symbols to A and O.
This was the only file i found where these changes made sense.
This is a proposed fix to #2252


menny (approved):
thanks.

Do you think we need the other keyboard layout (qwerty without symbols)?

1000283:
Thanks!
You mean [qwerty.xml](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/addons/languages/portuguese/pack/src/main/res/xml/qwerty.xml)?
Judging by the contents, i don't think so.

1000283:
Hi @menny, i see this was merged to master a month ago but it didn't seem to be in v1.10-r4? 
Or did i miss something?
Thanks.

Issue #1667, opened 2019-03-20
[Feature request/layout adjustment]


sunnaryt:
Hi, I'd just like to point out that you have redundant "," and "." keys on all dvorak layouts. This could be well remedied in my opinion by dropping the q and z keys into their position on the bottom, respectively. Gboard does this and this both saves space to fit keys onscreen and makes the keyboard easier to use in my experience. I dont want to use gboard anymore and this is the only thing holding me back from switching as I find that layout to be ideal.

Heres the dvorak layout gboard uses:
https://img.devrant.io/devrant/rant/c_541416_hWHaz.jpg

Soundtoxin:
There are also some missing things like semicolon to left of q, hyphen to right of s, etc. would be nice to copy the style of a 60% keyboard with dvorak for consistency.

PR #2814, opened 2021-04-12
Improve English Dvorak usability

danielrparks:
This pull request adds an English Dvorak layout with common symbols above the keys, similar to the existing QWERTY with symbols layout. It also adds a bottom row variant without the "." and "," keys, which are redundant when using Dvorak. This partially addresses the concerns in #1667.


danielrparks:
Hmm, it looks like the app-less-tests check has died due to an internal error and needs to be re-run.
That should happen automatically when I push the fix for the app-tests-shards test.

menny:
Thank you

danielrparks:
OK, that's fixed now.

menny (approved):

PR #2853, opened 2021-05-15
Moving summaries and descriptions from F-Droid repo

ArenaL5:
so that future updates will behave better.

There are some non-trivial tasks left to facilitate future updates:

- [ ] Upgrade Gradle to 7.0: requires changing `@Input` annotations under the `buildSrc` directory to `@InputFile`, `@InputFolder`, or nothing, and I don't know enough about Gradle to do it myself.
- [ ] Enforce use of JDK 11. Gradle has some tools for this: if I have only JDK 8 installed, I can make it auto-download JDK 11 by adding this block to `buildSrc/build.gradle`, but it will later fail because some tasks like `validatePlugins` are still running under JDK 8:

```gradle
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(11)
        vendor = JvmVendorSpec.ADOPTOPENJDK
    }
}
```


ArenaL5:
By the way, I'm going to be away for a few days, but I'm trying to initiate the process to update the Afrikaans app. After that PR is corrected and merged, I'll request update of every other language addon en masse.

menny:
This is amazing!
Thanks.

I see many references to f-droid.org in the description, can we change that? I don't want play-store descriptions to include f-droid

menny:
Regarding Java11, is there an option to run in a docker container? This is what we do in CI

ArenaL5:
I don't really know anything about docker images, but F-Droid already build all of their apps inside of throw-away virtual machines... which I think is basically it.

Setting Java 11 from Gradle is not really that important; it's mostly a nice touch for first-time collaborators (who might not have read CONTRIBUTING.MD). I suppose halting the build and instructing them to read CONTRIBUTING.MD would be even better. It _would_ remove a couple lines from the build recipe at F-Droid, but it's not like this is the only app on Java 11 anyways.

Good eye about references to F-Droid; all of them are only translations of the sentence _For more info see the f-droid.org [http://f-droid.org/forums/topic/anysoftkeyboard-language-packs-layouts-plans/#post-6408 forum topic]_, with a link to an outdated, auto-closed topic. They don't even belong on F-Droid. I'll push a new commit removing that sentence from every language.

ArenaL5:
Also it's really not that amazing. It's a copy-paste, a mass rename of files, and a link to the source in the commit description. I'm glad you like it, though. :relaxed:

ArenaL5:
Sorry I'm late, I've been very busy this couple of days.

I wrote a script to help me merge descriptions and summaries here with the ones on F-Droid. That's the content of the third commit on this PR. It should be easy to fix if you find something wrong.

I think I also removed any reference to F-Droid.

PR #2859, opened 2021-05-18
Add support for predefined system vibrations on Android Q and higher

danielrparks:
On my Pixel 3, the vibrations from AnySoftKeyboard feel much less comfortable and refined than Gboard or Swiftkey. That is because Gboard and Swiftkey support [predefined vibrations](https://developer.android.com/reference/android/os/VibrationEffect#EFFECT_CLICK) tuned to the hardware.

This pull request adds similar support to ASK.

Some notes:

- I have included a [test for the system vibrations feature](https://github.com/danielrparks/AnySoftKeyboard/blob/systemvibrate2/ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardPressEffectsTest.java#L288), but it depends on the getEffectId method which is only available starting with roboelectric 4.4, so it is currently disabled.
- I found a bug where ASK would play both the short- and long-press vibrations at once when long-pressing on a key with a popup keyboard. That is fixed in 969b1a3c390b4aed77fd3093e5c17582310bfd8e, which is part of this pull request.


ime/app/src/main/java/com/anysoftkeyboard/devicespecific/DeviceSpecificV26.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2021 Daniel Parks
```
menny, line 2: Please, change the copyright to AnySoftKeyboard. This is according to https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/CONTRIBUTING.md#license

ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardPressEffectsTest.java
```diff
@@ -280,6 +285,40 @@ public void testVibrateWhenEnabled() {
     }
 
     @Test
+    @Ignore("Requires roboelectric 4.4 or higher")
```
menny, line 288: On the latest `master` we have Robolectric 4.5.1

ime/app/src/main/res/values/settings_defaults_dont_translate.xml
```diff
@@ -3,6 +3,7 @@
     <!-- DO NOT TRANSLATE -->
     <!-- Default settings values -->
     <string name="settings_default_default_domain_text">.com</string>
+    <bool name="settings_default_use_system_vibration">false</bool>
```
menny, line 6: I assume that we want this to be `true` for API29+? If so, you can override the value for `values-v29`. (similar to what we do with [API 30](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/res/values-v30/settings_defaults_dont_translate.xml)

menny (changes requested):
This is amazing! Thank you for this.

I have a few comments.

ime/app/src/main/res/values/settings_defaults_dont_translate.xml
danielrparks: I actually meant for this to be `false` by default because support for the effects [is not mandatory](https://developer.android.com/reference/android/os/Vibrator#areAllEffectsSupported(int...)), but I looked more carefully and [it does fall back to a sane default](https://developer.android.com/reference/android/os/VibrationEffect#createPredefined(int)) if they are not supported.

I enabled it by default for 29+ and discovered via the tests that it wasn't correctly responding to night mode / low power, so I fixed that as well.

Definitely worth it!

danielrparks:
Do you know what happened with the tests? They all ran fine locally.

menny:
@danielrparks I've created https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2896 which fixes the merge conflicts. I've made a few changes since you created this PR, and didn't want to waste your time fixing them

danielrparks:
Thanks, looks good to me. (except for the CheckBoxPreference that was affected by the androidx migration)

menny:
yeah... missed that.
Please [review](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2896) and I'll merge

PR #2864, opened 2021-05-20
Fix backup failures

### LINKED ISSUE #2815: Backup failes due to missing directory
### Steps to reproduce
1. Start ASK
2. Select three dot menu
3. Select Backup
4. Click Backup

### Expected behaviour
Backup file should be created in /storage/emulated/0/Android/data/com.menny.android.anysoftkeyboard/files.

### Actual behaviour
Error message is displayed: Failed! Failed to backup your data due to Failed to of storage folder /storage/emulated/0/Android/data/com.menny.android.anysoftkeyboard/files

This is because the "com.menny.android.anysoftkeyboard" directory does not yet exist in /storage/emulated/0/Android/data.  If I manually create it, the backup works correctly.

Taking a quick look at the source code it looks like line 133 of AnyApplication.java is the likely culprit:
```
new File(externalFolder, "/Android/data/" + BuildConfig.APPLICATION_ID + "/files/"),
```
As it looks like it assumes all paths already exist above "files".

I'm not a java programmer, but doing a quick search, perhaps something like:
```
new File(externalFolder, "/Android/data/" + BuildConfig.APPLICATION_ID + "/files/").mkdirs(),
```
would solve the issue?

**Android OS version:**
LineageOS 17.1 or 18.1, no GAPPS or microG.

Muliptle: OnePlus 5,7,8t, Pixel 4, Samsung Galaxy Tab s5e

None

toolstack:
...due to missing external application directory.

Resolves #2815

### DIFF
--- a/ime/app/src/main/java/com/menny/android/anysoftkeyboard/AnyApplication.java
+++ b/ime/app/src/main/java/com/menny/android/anysoftkeyboard/AnyApplication.java
@@ -202,6 +202,13 @@ public class AnyApplication extends Application {
 
         mPublicNotices = new ArrayList<>(EasterEggs.create());
         mPublicNotices.addAll(Notices.create(this));
+        
+        // For Android 11 (maybe earlier?) we need to make sure the external application directory
+        // exists, to do so, just have the system get an empty file from it and Android will create
+        // it for us.  Likewise we have to do it here instead of in the getBackupFile function as
+        // we don't have the right context by that time.  We need this to write backups too and 
+        // with Android 11 we can no longer do this ourselves.
+        this.getExternalFilesDir("");
     }
 
     public List<PublicNotice> getPublicNotices() {


ime/app/src/main/java/com/menny/android/anysoftkeyboard/AnyApplication.java
```diff
@@ -202,6 +202,13 @@ public void onCreate() {
 
         mPublicNotices = new ArrayList<>(EasterEggs.create());
         mPublicNotices.addAll(Notices.create(this));
+        
+        // For Android 11 (maybe earlier?) we need to make sure the external application directory
+        // exists, to do so, just have the system get an empty file from it and Android will create
+        // it for us.  Likewise we have to do it here instead of in the getBackupFile function as
+        // we don't have the right context by that time.  We need this to write backups too and 
+        // with Android 11 we can no longer do this ourselves.
+        this.getExternalFilesDir("");
```
menny, line 211: We might not have the `android.permission.READ_EXTERNAL_STORAGE` granted at this time.

Can you share more information about this issue?

toolstack: Sure, according to the note in [this section ](https://developer.android.com/training/data-storage/app-specific#external-access-files) of the developers docs:

```Note: On Android 11 (API level 30) and higher, apps cannot create their own app-specific directory on external storage.```

So when a backup is run and it gets to the `getBackupFile()` call that runs:

```
return new File(
                new File(externalFolder, "/Android/data/" + BuildConfig.APPLICATION_ID + "/files/"),
                filename);
```
The APPLICATION_ID/files directory does not yet exist.

If you try and create it with `.mkdirs()` it fails as Android 11 will no longer let you create the APPLICATION_ID directory in '/Android/data'.   This then causes the backup to fail as it can't write to the non-existant directory.

Using `this.getExternalFilesDir("");` instead of `File(externalFolder, "/Android/data/" + BuildConfig.APPLICATION_ID + "/files/").mkdirs()` forces the system itself to create the directory during the class creation.

It could be moved in to the `getBackupFile()` function instead if `Context` was available but this does not appear to be the case.  I'm not that familiar with Java so perhaps there is a way to get a app context inside this function and then `getExternalFilesDir()` could be called only when the backup actually happens.

I did comple and test the new code, so permissions are available to do the creation (I suspect they aren't needed though as `getExternalFilesDir()` is using something other than the app context to do the directory creation otherwise it would fail just as `mkdirs()` does.

menny: thank you!

menny:
I see that your branch has conflicts. I created a new PR: https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2899
It also includes usages of Context.

Please review, and if fine we can merge the new PR and close this one. Your commit is still there.

toolstack:
Yep, looks fine to me.  I'll close this one.

PR #2890, opened 2021-06-10
Update README.md

Cyph0ZD:


menny (approved):

menny:
It fails because we use an RC version of build-tools (when we shouldn't). That version is gone from the servers.

menny:
thank you for the fix.

PR #2917, opened 2021-06-28
Pr delete recently used smileys

lubenard:
This PR add a button to clear recent emojis, under the image icon on the right when in emoji view.


api/src/main/java/com/anysoftkeyboard/api/KeyCodes.java
```diff
@@ -66,5 +66,7 @@
     public static final int PRE_PREPARED_TEXT_POPUP = -151;
     public static final int PRE_PREPARED_EMAILS_POPUP = -152;
 
+    public static final int DELETE_RECENT_USED_SMILEYS = -160;
```
menny, line 69: Why do we need this to be a key-code? I don't see how we would like this functionality everywhere, right?

I imagine this to be similar to how we react to the `Settings` icon: https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/master/ime/app/src/main/java/com/anysoftkeyboard/quicktextkeys/ui/FrameKeyboardViewClickListener.java#L32 - have the key-press directly call `clearHistory()`

menny (changes requested):

api/src/main/java/com/anysoftkeyboard/api/KeyCodes.java
lubenard: I agree, but settings also seemed to be a key-code. https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/b68849e7a8f9714c1ff9d91acf49b7385f4d0d7b/api/src/main/java/com/anysoftkeyboard/api/KeyCodes.java#L39
This is why i added it to the list of key-codes.

ime/app/src/main/java/com/anysoftkeyboard/quicktextkeys/QuickKeyHistoryRecords.java
```diff
@@ -79,7 +79,18 @@ private static String encodeForOldDevices(@NonNull List<HistoryKey> outputSet) {
...
+        mLoadedKeys.clear();
+        // For a unknown reason, we cannot have 0 history emoji...
+        mLoadedKeys.add(new HistoryKey(DEFAULT_EMOJI, DEFAULT_EMOJI));
+        final String encodedHistory = encodeForOldDevices(mLoadedKeys);
+        mRxPref.set(encodedHistory);
+    }
+
     public List<HistoryKey> getCurrentHistory() {
+        if (mLoadedKeys.size() == 0)
+            // For a unknown reason, we cannot have 0 history emoji...
```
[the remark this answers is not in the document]
lubenard: I want to change this behavior. Is it okay ?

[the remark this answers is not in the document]
menny: Sure

api/src/main/java/com/anysoftkeyboard/api/KeyCodes.java
menny: that's a different settings. The settings icon in the emoji layout navigates directly to the quick-text (emoji) settings. It does not use the `onKey` mechanism.

I suggest calling `clearHistroy()` directly there instead of calling `onKey` with the new key-code

lubenard: Okay, il will fix this tonigh.

lubenard: It seems not to be working when not using ```onKey```. 

It seems be linked with the ```getQuickKeyHistoryRecords()``` that i am able to fetch in the ```AnySoftKeyboard.java``` class, but not in ```FrameKeyboardViewClickListener.java```

I tried to use ```new QuickKeyHistoryRecords(AnyApplication.prefs(v.getContext())).clearHistory();``` instead of onKey, but it's not deleting anything. Weird

menny (approved):

menny:
thank you

PR #2924, opened 2021-07-02
Using suggestion close icon from theme

JohnOberhauser:
### Issue
The keyboard is not using the close icon from the theme from `<attr name="suggestionCloseImage" format="reference"/>`

### Description
- Created an interface to update the close icon image in `CancelSuggestionsAction` that gets triggered when the `CandidateView` grabs the `suggestionCloseImage` from the theme.

The icon in question:
close_icon

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
@@ -5,6 +5,7 @@ import android.content.res.Configuration;
 import android.graphics.drawable.Drawable;
 import android.os.SystemClock;
 import android.text.TextUtils;
+import android.util.Log;
 import android.util.SparseBooleanArray;
 import android.view.KeyEvent;
 import android.view.LayoutInflater;
@@ -15,6 +16,8 @@ import android.view.animation.AnimationUtils;
 import android.view.inputmethod.CompletionInfo;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
+import android.widget.ImageView;
+
 import androidx.annotation.CallSuper;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
```diff
@@ -1461,6 +1468,17 @@ public void onRemoved() {
             mRootView.removeCallbacks(mReHideTextAction);
         }
 
+        public void setCloseIconDrawable(Drawable closeIcon) {
+            mCloseIcon = closeIcon;
+        }
+
+        private void setCloseIcon() {
+            ImageView closeIcon = (ImageView) mRootView.findViewById(R.id.close_suggestions_strip_icon);
+            if (mCloseIcon != null) {
```
menny, line 1477: why would it be `null`?

ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
```diff
@@ -1461,6 +1468,17 @@ public void onRemoved() {
             mRootView.removeCallbacks(mReHideTextAction);
         }
 
+        public void setCloseIconDrawable(Drawable closeIcon) {
+            mCloseIcon = closeIcon;
```
menny, line 1472: Can you apply this `drawable` to the `ImageView` here?

ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
JohnOberhauser: I cleaned up some stuff.  Can you review again?

ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardSuggestions.java
JohnOberhauser: removed this method in cleanup

menny (changes requested):
This looks good, but I want to see themes that use this. Can you add light/dark versions? (I guess the current icon is for _dark_ themes)

menny:
continuing work in https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/2998

Issue #2846, opened 2021-05-10
 Keyboard appending an extra space while glide typing


AngmarLord:
### Steps to reproduce
1. Enable gesture typing
2. Use only gesture typing to type
3. On some occasions, especially if one has picked a recommended  word, or even otherwise, such as the extra space before the word,  "word", or after the comma before "word" in this sentence, an extra space is inserted by keyboard. 

### Expected behaviour
No extra  space (happened before "space" here, too) ought to be inserted  between words (another one before "between" just before) . 

It's difficult to see here, since the editor is  deleting the extra space. I inserted several here, but in the actual presentation of this issue, the extra space has been deleted. 

### Actual behaviour
Extra space is inserted between some words, randomly. 

**Android OS version:**
9

**Device manufacturer and model:**

**List of installed add-ons (like languages, or themes):**
Default English keyboard, no special theme

sonofkraken:
I just upgraded to the newest release (1.10.1786) and can confirm the same is happening to me. It seems to happen only when selecting a recommended word. The space is inserted and I have to backspace to add punctuation.

maximilianovermeyer:
Possible duplicate of #2817, maybe linked to #2813 and #2841

eatmeek:
I'm  having the same issue here. It happens during gesture typing  and selecting a word suggestion - any punctuation after is preceded by a space.

Issue #2841, opened 2021-05-07
space after "(" in gesture typing


paradise4all:
### Steps to reproduce
1. write: (
2. than gesture type
3.

### Expected behaviour
the next word should not create a space: (example)

### Actual behaviour
it crates a space: ( example)

**Android OS version:**
10
**Device manufacturer and model:**
Motorola g7power
**List of installed add-ons (like languages, or themes):**
german & english

maximilianovermeyer:
Possible duplicate of #2813

ej159:
This seems to still be an issue ( definitely)

kurahaupo:
Which punctuation gets spaces is locale-specific and context-specific; it would be helpful for these to be configurable on a per-locale basis:

* spaces inside parentheses;
* spaces inside quotation marks;
* spaces around long dashes;
* spaces around short dashes;
* spaces after periods, commas; colons, semicolons, exclamation marks, question marks, etc
* removing spaces between consecutive punctuation marks

It might also be useful to have a pseudo-locale called `code` that's for writing software rather than any natural language, that avoids any adjustments to what the user explicitly types.

PR #2943, opened 2021-07-29
Fix double call to pickSuggestionManually when swiping after correction

### LINKED ISSUE #2813: Punctuation and spaces: ( open parenthesis or bracket), " speech marks" and @ symbol.org
### Steps to reproduce
1. Type (
2. Gesture a word... e.g. Example
3. Type )

### Expected behaviour
(Example)

### Actual behaviour
( Example)

**Android OS version:**
Nougat

**ASK version**
1.10.r3 ( awaiting on FDroid)

OnePlus 3T

### LINKED ISSUE #2817: Double space with gesture typing on v1.10-r4
### Steps to reproduce
1. Use gesture typing and select a word from the suggested ones
2. Write another word with gesture typing

### Expected behaviour
There should be a word, a space and a word

### Actual behaviour
There are a double space between the two words

**Android OS version:** Android 11

**Device manufacturer and model:** Xiaomi Redmi Note 5

**List of installed add-ons (like languages, or themes):** None

danielrparks:
There is currently an issue with the way ASK handles word selection during gesture typing such that if the user selects an option from the suggestions bar after swiping a word, ASK will internally select the suggestion _again_ when the user begins swiping the next word. This change ensures that pickSuggestionManually will only be called once per word during gesture typing.

This fixes #2817 and #2846, but does not fix #2813 or #2841. I believe those are (at least partly) separate issues.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardWithGestureTyping.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardWithGestureTyping.java
@@ -374,7 +374,6 @@ public abstract class AnySoftKeyboardWithGestureTyping extends AnySoftKeyboardWi
     private void confirmLastGesture(boolean withAutoSpace) {
         if (mJustPerformedGesture) {
             pickSuggestionManually(0, getCurrentComposedWord().getTypedWord(), withAutoSpace);
-            mJustPerformedGesture = false;
         }
     }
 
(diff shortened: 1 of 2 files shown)


menny:
This is great. Can you add a unit-test for this?

danielrparks:
@menny Ok, I added the unit test. I tested it on current master, and it correctly fails unless 970ea67 is applied.

menny (approved):

menny:
Thank you!

PR #3019, opened 2021-10-10
Fix pt-BR translation for string setup_wizard_step_download_details

vitor-ramos:
I noticed while trying the app that the wording didn't make sense. Fixed to match the english version. I also verified that in values-pt (Portugal) the translation is correct.


vitor-ramos:
Done, I also voted the string double_space_to_period which has fixed a typo.

PR #3040, opened 2021-10-24
refactor(ossturkish): swap accented popup characters for symbols in Turkish language pack

### LINKED ISSUE #3038: Meaningful popup characters for the Turkish language pack
Turkish language pack has tons of popup characters that don't make the slightest sense (this is coming from a native speaker). I'm proposing to get rid of them and instead add symbols that are used frequently (such as exclamation and question marks and so on). I'm willing to create a PR for this if the proposal is accepted.

MusKaya:
Turkish language only has a few accented characters. Remove all unnecessary
accented characters from popups and instead add symbols like ' / ? !

Also swap the Turkish keyboard with more accented characters with an alternate
that has English QWERTY layout but the accented characters that are specific to
Turkish language are baked into the popup characters as the first popup character
of the related characters. Namely, ı is the first popup character of i, ü of u, ş of s,
ç of c and ö of o.

Closes #3038


MusKaya:
hey, I added the missing escape character before question marks, please re-run, thanks!

menny (approved):

menny:
thank you!

Issue #3031, opened 2021-10-17
Very few kaomojis available to use


Netherquark:
### Steps to reproduce
1. Open competing keyboard (eg GBoard)
2. Check Kaomoji section
3. Can be observed to have more kaomojis than AnySoftKeyboard.  

### Expected behaviour
All of [these](https://photos.google.com/share/AF1QipOFUiRc_A4RxyVMNAmeQE8fJR7nbkumoC1NfitJzQxfpIU95dFsn5v2wC3bTJfaXw?key=Y3BZSEFXcUttTGszVno3Tndodm9neEFmYUhxMXpR) emojis are available to the user.

### Actual behaviour
A selective few kaomojis are available

**Android OS version:**
Android 11 R

**Device manufacturer and model:**
Xiaomi Redmi Note 9 Pro (curtana/miatoll)

**List of installed add-ons (like languages, or themes):**
Language: Only English
Theme: AOSP Dark
That's it? stock apart from that

yagueto:
+1 for this, would like to see more as well.

It could be great to allow users to add their own kaomoji as well (without recompiling the app)

Netherquark:
[This](https://www.androidpolice.com/2019/01/15/gboard-v7-9-prepares-new-emoticons-adds-many-new-languages-and-more-apk-teardown/) might be relevant. Apart from that, a quick search leads to many kaomoji websites such as [this](http://kaomoji.ru/en/), [this](https://kaomojihub.com/), and more.

christofjs:
I can work on this. I'll try to get a pull request up in the next couple days. Let me know if there's any more input before I start (e.g. is there such thing as too many). I'll organize the xml by emotion/creature via comments, but add them all to the current kaomoji tab for now.

Netherquark:
that's awesome, just adding them to the current kaomoji tab should be a good initial start imo

Netherquark:
Sooo @christofjs is this fixed?

christofjs:
Yes and no @Netherquark ... got them all added and have a pull request in, but the build is failing for an unknown and I would have to imagine an unrelated reason. Just asked Menny to take a look, but I can't think of anything more that I can do.

Netherquark:
Thats great!

menny:
This is merged and will be available in the next release. Thank you all

PR #3100, opened 2021-12-07
categorized and added more kaomoji

christofjs:
I didn't change any of the lines that were in there, so those kaomoji still end in `\u0020`. I'm not sure why that's there. I couldn't add it similarly to the new additions without it appearing as part of the kaomoji during local testing. Definitely something that someone more familiar with the code should look at. Maybe I'm missing something and it does need to be added.

Responds to issue #3031 .

### DIFF
--- a/ime/app/src/main/res/xml/popup_kaomoji.xml
+++ b/ime/app/src/main/res/xml/popup_kaomoji.xml
@@ -5,128 +5,128 @@
         android:keyHeight="@integer/key_normal_height">
     <Row android:rowEdgeFlags="top">
         <!--Joy-->
-        <Key android:keyLabel="(* ^ ω ^)" android:keyOutputText="(* ^ ω ^)\u0020" />
-        <Key android:keyLabel="(´ ∀ ` *)" android:keyOutputText="(´ ∀ ` *)\u0020" />
-        <Key android:keyLabel="٩(◕‿◕｡)۶" android:keyOutputText="٩(◕‿◕｡)۶\u0020" />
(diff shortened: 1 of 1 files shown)


menny:
This is pretty cool. Thanks.

You can add tags to the kaomoji:  
`ask​:​tags​=​"​happy,joy,kjoy"​` for example

christofjs:
@menny, is this good? Not sure why it's failing, I didn't change anything significant. It's saying failed to build for a Gradle version mismatch, running on 7 rather than 8.

ime/app/src/main/res/xml/popup_kaomoji.xml
```diff
@@ -1,49 +1,132 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
-          android:keyWidth="15%p"
-          android:keyHeight="@integer/key_normal_height">
+        android:keyWidth="15%p"
```
menny, line 4: missing namespace `ask`.
add `xmlns:ask="http://schemas.android.com/apk/res-auto"`

menny (changes requested):

ime/app/src/main/res/xml/popup_kaomoji.xml
christofjs: Hopefully got it in there right.

menny: seems right

menny (approved):

Issue #2915, opened 2021-06-25
Recent change in Spanish popups


JohnVeness:
Hi, I use ASK in English and Spanish (the latter mainly for Duolingo), obtained from F-Droid.

Until recently, the long-press of keys on the Spanish keyboard showed just one alternative, which was the version of the letter with the accent. So long-pressing e would give you a choice only of é, and long-pressing i would give you a choice only of í.

This was fine for Duolingo use as that is pretty much you only accent you need, and you don't need numbers. Indeed, even if I had dictionaries turned off, I usually switched to the Spanish keyboard in Duolingo, just for the ease of typing á é í etc., compared to how to do the same in the English keyboard.

Recently, I think due to an update, now long-pressing e gives me a choice of éèëęē3€³₃ê, with 3 being the default, and long-pressing i gives me a choice of *ιīłïîíì8, with 8 being the default.

This means the Spanish popups are the same as the English ones. Maybe that is useful to many people, but for my particular use case this is a downgrade in usability!

Is it possible to restore the old behaviour, or at least making the á é í etc accent the default on those keys?

By the way, I say I think this was caused by an update: F-Droid showed me an update for "Israel 64 BDay Theme" a few days ago, which I don't actually have installed. However I am aware that F-Droid often shows the wrong app names for ASK and its add-ons for some reason, so I accepted the update. I don't know if this was really an update for the main ASK app or for the Spanish pack.

Thanks in advance!

dmlls:
As an example, the German keyboards do correctly default to the Umlaut instead of a number on long-press:

![IMG_20220202_143526](https://user-images.githubusercontent.com/22967053/152164391-ff58cb13-0d4c-413a-a74d-c9b3a800dd0e.jpg)

dmlls:
@JohnVeness this issue can now be closed, since the PR has been merged 🙂

JohnVeness:
Thank you! I look forward to trying it.

PR #3231, opened 2022-02-02
Spanish: prioritize accents over numbers in popups

dmlls:
See https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/2915.

In Spanish, graphic accents ("[acentos gráficos](https://es.wikipedia.org/wiki/Acento_gr%C3%A1fico)" or "tildes") are extremely common. These accents appear only on the vowels (áéíóú). Right now, the default popup keys for these letters on long-press are numbers, which makes it cumbersome to add accents.

This PR prioritizes accents over numbers on the letters "aeiou", since (at least in Spanish) it is more common to long-press to add an accent than to add a number (plus generally there are more suitable options to add numbers: number row, symbols/numeric keyboard).


menny (approved):

menny:
thank you!

PR #3235, opened 2022-02-06
Bump AndroidX Navigation 2.4.0

Unpublished:
Fixes opening ASK settings from launcher on master in release build.

https://developer.android.com/jetpack/androidx/releases/navigation#2.4.0
Relevant bug fix (since 2.4.0-alpha04):
> Add ProGuard rules for @Navigator.Name, fixing issues when using R8 3.1 full mode.

See https://issuetracker.google.com/issues/191654433

Also remove deprecated options from gradle.properties.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,6 +5,4 @@ org.gradle.caching=true
 android.useAndroidX=true
 android.testConfig.useRelativePath=true
 android.enableJetifier=true
-android.enableD8=true
-android.enableR8=true
 android.enableR8.fullMode=true
--- a/gradle/root_all_projects_ext.gradle
+++ b/gradle/root_all_projects_ext.gradle
@@ -4,7 +4,7 @@ allprojects {
         robolectricVersion = '4.7.3'
 
         androidxCore = '1.3.2'
-        androidxNavigation = '2.3.5'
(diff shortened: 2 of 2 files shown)


gradle.properties
```diff
@@ -5,6 +5,4 @@ org.gradle.caching=true
 android.useAndroidX=true
 android.testConfig.useRelativePath=true
 android.enableJetifier=true
-android.enableD8=true
```
menny, line 8: @Unpublished please add a note here why you removed these

menny (changes requested):
thanks!

gradle.properties
Unpublished:
```
WARNING:The option 'android.enableD8' is deprecated.
The current default is 'true'.
It was removed in version 7.0 of the Android Gradle plugin.
For more details, see https://d.android.com/r/studio-ui/d8-overview.html.
WARNING:The option 'android.enableR8' is deprecated.
It was removed in version 7.0 of the Android Gradle plugin.
Please remove it from `gradle.properties`.
```
Per above these options don't have any effect since they're defaults now.

menny (approved):

PR #3284, opened 2022-03-17
addition of new german keyboard

Pippo8291:
* added keyboard with ö,ä,ü keys but without ß


menny:
Thank you. What is the purpose of this keyboard? Is this a personal preference or a known variant?

Pippo8291:
thanks for your reply. To be honest it may be a kind of personal flavour, but i found no keyboard that shows the german ü, ä and ö key but hiding the ß key, which was the keyboard within my old device and which is the default setup of german keyboards for the computer. 
maybe i should change the name displayed within the spacebar.

menny (approved):

Pippo8291:
does a re-run fix this or do i have to add sth for the linters check?

menny:
well, look at that..

Pippo8291:
That was really quick. Thanks. Sign up is only possible with google play right?

menny:
yes, play store only

PR #3319, opened 2022-04-20
Set the total keywidth of bottom row keys to 100

mhdzli:
Change the default keyWidth and match the most keys attribute.
The total kewidth of the bottom row was 105% after adding change layout button (sorry my mistake) and a part of the Enter key was out of screen. change the keywidth of the bottom row keys to be 100% totally.

### DIFF
--- a/addons/languages/persian/pack/src/main/res/xml/persian_qwerty.xml
+++ b/addons/languages/persian/pack/src/main/res/xml/persian_qwerty.xml
@@ -2,7 +2,7 @@
 
 <Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
           xmlns:ask="http://schemas.android.com/apk/res-auto"
-          android:keyWidth="10%p"
+          android:keyWidth="9.09%p"
       >
     <!-- "android:keyWidth" specify the default width of a key. In this example, 10% of parent (the whole keyboard) -->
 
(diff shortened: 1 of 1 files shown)


menny (approved):

menny:
thanks for noticing

mhdzli:
@menny 
Is it possible to add label for popup characters? I want to add [Left-To-Right Embedding](https://unicode-table.com/en/202A/), [Right-To-Left Embedding](https://unicode-table.com/en/202B/) and [Pop Directional Formatting](https://unicode-table.com/en/202C/) as popup characters on space key. They don't have any visual representation to show up as label on popup characters, but they are necessary for writing English text in Persian paragraphs and Persian text in English paragraphs.

menny:
Well, you could use `android:keyLabel="text"` attribute in the `Key` node. That should allow you to show whatever you want on the key, but will still output the codes defined

mhdzli:
@menny 

`android:keyLabel="text"` attribute sets the label for the key but is there any way to set the labels for popup characters?

And how i can subscribe and join the BETA channel of Persian addon?

menny:
the `popupCharacters` attribute is a convenience API: it is actually being expended to `android:popupKeyboard` and pointing to an artificial XML layout.

You can replace the `popupCharacters` with `android:popupKeyboard` that points to a full-fledge keyboard layout.
For example, in the 16 keys English layout, we have popup for each key: https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/main/addons/languages/english/pack/src/main/res/xml/eng_16keys.xml#L10 (in this case, we see that the `ABC` key has a `popupKeyboard` attribute pointing to `@xml/popup_16keys_abc` which is https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/main/addons/languages/english/pack/src/main/res/xml/popup_16keys_abc.xml

menny:
About the Persian pack BETA, I need to fix the channel to be public. Let me do that and get back to you

PR #3325, opened 2022-04-22
add new verbs

154pinkchairs:
added some feminine forms and less common verbs


menny:
There is a better, and more efficient way to add words: provide a pre-build word list. For example, in English we have this: https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/main/addons/languages/english/pack/dictionary/prebuilt/additionals.xml

PR #3356, opened 2022-05-11
Using character number rather than character

menny:
this will help solving issues where FS/encoding changes these


GoRaN909:
Thank you so much @menny I can confirm that the Error is fixed by this PR!

PR #3360, opened 2022-05-13
Dictionary and keyboard layout update

sslaia:
Hi @menny I have updated Nias dictionary to version 2 and the keyboard layouts including adding a wiki keyboard, which is in demand among Nias language activists. Thanks.


menny (approved):

menny:
thanks

PR #3415, opened 2022-06-18
Gesture typing respects the initial state of the shift key

JellyWX:
With auto-capitalise off, the shift key would disengage when
gesture-typing and the word would be full lowercase. This ensures that
if shift is active when the gesture starts, then the word will be
shifted.

### DIFF
--- a/ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardGestureTypingTest.java
+++ b/ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardGestureTypingTest.java
@@ -139,6 +139,20 @@ public class AnySoftKeyboardGestureTypingTest extends AnySoftKeyboardBaseTest {
     }
 
     @Test
+    public void testOutputCapitalisedOnShiftLocked() {
+        mAnySoftKeyboardUnderTest.simulateKeyPress(KeyCodes.SHIFT_LOCK);
+        simulateGestureProcess("hello");
+        Assert.assertEquals("HELLO", mAnySoftKeyboardUnderTest.getCurrentInputConnectionText());
+    }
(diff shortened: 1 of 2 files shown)


menny (changes requested):
Thanks. Can you add a test for this?

JellyWX:
:+1:

ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardGestureTypingTest.java
```diff
@@ -138,6 +138,20 @@ public void testOutputPrimarySuggestionOnGestureDone() {
         Assert.assertEquals("hello", mAnySoftKeyboardUnderTest.getCurrentInputConnectionText());
     }
 
+    @Test
+    public void testOutputCapitalisedOnShiftLocked() {
```
menny, line 142: Nice! For completion, can you do a test for not-shifted (although it is covered by other unit tests)?

menny (changes requested):

ime/app/src/test/java/com/anysoftkeyboard/ime/AnySoftKeyboardGestureTypingTest.java
JellyWX: This would be identical to the test above (`testOutputPrimarySuggestionOnGestureDone`). I can rename that test if you like?

menny: no, this is good. I missed that.

menny (approved):

menny:
thanks

PR #3416, opened 2022-06-18
Backspace after gesture clears word

JellyWX:
Similar to Gboard, pressing backspace immediately after gesture input
will clear the word you gestured.

Good for correcting mistaken gesture input.

Marked draft since I haven't been able to run the tests yet

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardWithGestureTyping.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardWithGestureTyping.java
@@ -4,6 +4,7 @@ import android.graphics.drawable.Drawable;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


menny:
Thank you. I have a suggestion for a different UX: backspace will still delete one character at a time, but after a gesture we will show an icon on the suggestion bar (next to the X) that clears the gesture output. 
This will keep the backspace behavior normal and predictable, but also allows clearing the whole gesture

JellyWX:
Yes, that sounds good. I will undo that last commit with the unit test changes.

menny:
okay, so `KeyboardViewContainerView.StripActionProvider` is the thing that adds the icon on the suggestions bar.
For example, this is controller for the Clipboard paste action: https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/3dfd69a4eb734764490abb838d8646b1253a40c6/ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardClipboard.java#L78

menny:
I guess you can add/remove/hide/etc that in `AnySoftKeyboardWithGestureTyping`

JellyWX:
Menny, I've done an implementation here that works at least for me. I'm not an Android developer or Java developer, so I hope this code is sensible :) 

I'll work on the unit-tests now

ime/app/src/main/java/com/anysoftkeyboard/ime/AnySoftKeyboardWithGestureTyping.java
```diff
@@ -374,6 +382,7 @@ private void confirmLastGesture(boolean withAutoSpace) {
         if (mJustPerformedGesture) {
             pickSuggestionManually(0, getCurrentComposedWord().getTypedWord(), withAutoSpace);
         }
+        getInputViewContainer().removeStripAction(mClearLastGestureAction);
```
menny, line 385: The action should also be removed when `onFinishInput()` event is called

menny (commented):
This looks great!
How is the UX? Do you feel it makes sense like that?

I have a suggestion, let me know what are your thoughts:
 - add the action in `onStartInputView`, if gesture is enabled
 - remove the action in `onFinishInputView`
 - Call `mRoot.setVisibility` with `View.VISIBLE` or `View.GONE` when needed to show the action

JellyWX:
It's okay: it could do with hiding the suggestion "x" button I think.

There's also a bug in the current version, where backspace doesn't hide the clear gesture button. I'll try and address these when I get the chance.

-------- Original Message --------

menny:
I would like to see this in. How can I help?

JellyWX:
I'm not sure exactly what the test failures are here: can you give some guidance on what needs sorting here?

menny:
Okay... I found the issue: the tests were super fragile there: counting children rather than specifically looking for a type. See: https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/3416/commits/c1fdc13be0f0e40b9bc64f2214ba769cb589660d

menny (approved):

menny:
Thank you!

JellyWX:
Thanks :)

JellyWX:
Great, I'll have a look at this later

-------- Original Message --------

PR #3426, opened 2022-06-26
Handle keyHeight if given as dimension/fraction

djpohly:
When running aapt on keyboard XML files, it allows android:keyHeight to be a dimension or fraction as well as a reference to one of the height codes. Previously, values like "20dp" or "10%p" were ignored and replaced with the default key height. This commit allows layouts to specify precise dimensions if desired.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/keyboards/Keyboard.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/keyboards/Keyboard.java
@@ -673,6 +673,8 @@ public abstract class Keyboard {
         if (value == null) {
             // means that it was not provided. So I take my mParent's
             return defaultHeightCode;
+        } else if (value.type == TypedValue.TYPE_DIMENSION) {
+            return a.getDimensionPixelOffset(remoteIndex, defaultHeightCode);
         } else if (value.type >= TypedValue.TYPE_FIRST_INT
                 && value.type <= TypedValue.TYPE_LAST_INT
                 && value.data <= 0
--- a/ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardSupport.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardSupport.java
@@ -81,9 +81,7 @@ public class KeyboardSupport {
                 height = keyboardDimens.getLargeKeyHeight();
                 break;
             default:
-                height = heightCode >= 0
-                    ? heightCode
-                    : keyboardDimens.getNormalKeyHeight();
+                height = heightCode >= 0 ? heightCode : keyboardDimens.getNormalKeyHeight();
                 break;
         }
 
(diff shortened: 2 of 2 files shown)


ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardSupport.java
```diff
@@ -71,17 +71,17 @@ public static int getKeyHeightFromHeightCode(
...
                 break;
             case -2:
                 height = keyboardDimens.getSmallKeyHeight();
                 break;
             case -3:
                 height = keyboardDimens.getLargeKeyHeight();
                 break;
-            default: // -1
-                height = keyboardDimens.getNormalKeyHeight();
+            default:
```
menny, line 83: Previously, if the value of `heightCode` was invalid, we would show the normal-key-height. Now, let's say `-7` comes in, what will happen?

menny:
May I ask why you want this change?

ime/app/src/main/java/com/anysoftkeyboard/keyboards/KeyboardSupport.java
djpohly: Good catch. I'll fix this.

djpohly:
I've been fiddling with popupKeyboards in a custom layout, and I wanted to be able to put a very short "default" key at the bottom of a popup so that a long-press and immediate release would reliably trigger this key, but without sacrificing an entire row's worth of screen real estate.

All I really need is dimensions.  I'm less happy with the fraction code, and I wouldn't mind removing it.  (I had given it a shot because aapt seemed to indicate fractions are valid, and because of the comment in the API indicating ["in pixels or percentage of display width"](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/main/api/src/main/res/values/keyboard_layout_api.xml#L23).)

ime/app/src/main/java/com/anysoftkeyboard/keyboards/Keyboard.java
```diff
@@ -498,7 +498,7 @@ private void setDataFromTypedArray(
                                     a, remoteIndex, mKeyboard.mDisplayWidth, parent.defaultWidth);
                     break;
                 case android.R.attr.keyHeight:
-                    int heightCode = getKeyHeightCode(a, remoteIndex, parent.defaultHeightCode);
+                    int heightCode = getKeyHeightCode(a, remoteIndex, mKeyboard.mDisplayWidth, parent.defaultHeightCode);
```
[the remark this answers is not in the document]
djpohly: It seemed to be the only base dimension that was easily accessible.  I agree it doesn't make a ton of sense, and I wouldn't mind removing the fraction stuff altogether.

djpohly:
I've removed the semi-hacky fraction support and made sure that invalid `heightCode` values will default to normal height.

menny (dismissed):

menny:
@djpohly I pushed a commit that fixes a formatting issue, but CI still failed.
Seems to be a GitHub Actions regresion. So, I pushed this: https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/3434

I hope it will pass now

menny (approved):

PR #3543, opened 2022-11-17
Fix MIME type for media insertion

### LINKED ISSUE #3542: Picture insertion does not remember mime type or show preview
### Steps to reproduce
1. tap emoji button
2. tap picture button
3. choose a picture
4. tap back on the text box as it says to (why do I have to do this?)

### Expected behaviour

Image should appear in the text box, sending the image should result in an image in my app.  This is what happens if I use eg gboard

### Actual behaviour

Big grey square with question mark appears in the text box, sending it results in a file transfer but my app doesn't know the type of the file and cannot view it.

singpolyma:
When we proxy we lose the MIME type information, so set the file extension based on the MIME type so that contentResolver.getType() will still work on the new URI, this fixes image preview in the text box etc.

Fixes #3542

### DIFF
--- a/ime/fileprovider/src/main/java/com/anysoftkeyboard/fileprovider/LocalProxy.java
+++ b/ime/fileprovider/src/main/java/com/anysoftkeyboard/fileprovider/LocalProxy.java
@@ -4,6 +4,7 @@ import static androidx.core.content.FileProvider.getUriForFile;
 
 import android.content.Context;
 import android.net.Uri;
+import android.webkit.MimeTypeMap;
 import androidx.annotation.NonNull;
 import com.anysoftkeyboard.base.utils.Logger;
 import com.anysoftkeyboard.rx.RxSchedulers;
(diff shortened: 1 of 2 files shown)


menny:
Can you add a screenshot before and after the fix?

singpolyma:
# before
screenshot_before

# after
screenshot_after

menny:
@singpolyma have you looked at the error in the unit-test? LocalProxyTest

menny:
okay, I was able to make some changes. @singpolyma see https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/3543/commits/4ef6348fc19222bd3e76adbe8a954299e8553423

menny:
@singpolyma can you check if this fix (which I pushed to your branch, so you can `git pull` it) works and fixes your issue?

singpolyma:
@menny sorry for the slow reply, I have tested with your changes and it still seems to work for me

menny:
Thanks for checking. I'll merge this

menny:
Thank you for the fix

menny (approved):

Issue #3003, opened 2021-09-23
Key alphabets missing in Hindi/Devnagari keyboard


AngmarLord:
Thanks for finally publishing the Hindi keyboard for AnySoftKeyboard.

Some key letters are missing in the keyboard, though.

The "h" sounding **ह**, using which one would actually write Hindi in Devnagari is missing in the keyboard layout.

Similarly,  the "v" sounding **व** is missing as well.

Kindly add these and any other missing characters in layout, without these key letters, this language keyboard would be difficult to use. 

 ### Steps to reproduce
1. Install Hindi language pack from Play store
2. Switch keyboard from English to Hindi

### Expected behaviour
All characters in Devanagari keyboard should be available. 

### Actual behaviour
Several characters are missing. 

**Android OS version:**
9

**List of installed add-ons (like languages, or themes):**
No specific add-ons or themes. Default theme. Alt number layout with English.

PR #3558, opened 2022-12-01
Added missing characters for Hindi language

Raj9039852537:
Consider updating Hindi add-on so it can become usable


menny:
Thank you. Can you add context and reasoning? All I see is that you removed characters and added others

Raj9039852537:
In normal inscription layout also these characters are there but with ask it is impossible to type some words like "नही" "हाँ" "लेना" etc due to these missing characters
Here is the physical keyboard image clearly showing missing symbols from ask
[Hindi keyboard image from will feature for reference](https://upload.wikimedia.org/wikipedia/commons/9/99/Devanagari_INSCRIPT_Keyboard.JPG) 

Also every other keyboard that support Hindi have these characters included.

It is also reported [Here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/3003)

Raj9039852537:
Also he character I changed (not removed) are already there with shift and long press action which is just normal for those characters as per the given Wikimedia image above

menny:
thank you for the context

menny (approved):

Raj9039852537:
Thanks 🤝

PR #3593, opened 2022-12-29
Added BDS Modern Bulgarian keyboard layout

Bugzey:
Added a new keyboard layout - Bulgarian BDS Modern - that removes unused symbols from the Bulgarian national standard (Български държавен стандарт) layout and optimizes wasted space that is present in the regular BDS layout.

I have also added number hints and numbers as popup characters.

image


menny:
When you say _modern_ does this hint that it should be the default layout?

menny:
@Bugzey ping

PR #3618, opened 2023-01-20
Introducing meaningful gesture detection. Fixes #3615

### LINKED ISSUE #3615: Very quick and short gestures should not be registered as gesture typing
For example, quickly slide your finger from P to O. This will result in a gesture typing (pop) where this is clearly just a finger slide from on key to the neighboring key

menny:
Discards gestures that are either too short or too quick.


menny:
closing because https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/3619

PR #3625, opened 2023-01-26
add ŷ to popup

puffinux:
this would add the ŷ (y with ^ diacritic) symbol to the english qwerty with symbols keyboard, when you long press y, added at the end of the popup options because they appear to be arranged in UTF8 symbol order and ŷ comes after the other current popup options


puffinux:
i'm asking for this character specifically because it would allow typing in Welsh / Cymraeg, a language not currently supported as a separate keyboard. the standard english qwerty keyboard works fine but requires a few additional ï ŵ ŷ î etc and so the qwerty with symbols keyboard works fine, except it lacks the ŷ character (current popup options are 6ýÿ, this merge would increase to 6ýÿŷ)

puffinux:
i'd be happy to create a separate Cymraeg keyboard (most likely would be very similar to (qwerty + some symbols) but adding a whole language seems like a hassle and i've no clue how to create the dictionary etc, so this is by far the easiest workaround for now

menny (approved):

menny:
This change makes sense. Thank you.

PR #3631, opened 2023-02-03
Russian with symbols keyboard

dgalaktionov:
~~A new keyboard akin~~ Add longpress symbols to the default Russian layout, making it similar to English `qwerty_with_symbols`, following the same layout for symbols where possible.

No other such keyboard exists currently for Russian. It should be very handy for people like me that enjoy a layout with symbol popups, and find themselves frequently switching between English and Russian layouts.


menny:
Is this a similar layout to the default Russian keyboard?

dgalaktionov:
@menny yes, it is basically the same layout, with more symbol popups added. There is only one small difference: the order of popups over the "е" at the top is now "5ё" instead of "ё5". I find the default very annoying because it broke the consistency for quick number typing, while most people don't even bother writing ё in informal contexts.

menny:
It seems to me that we want to change the default layout to include your changes, rather than create a new one. Makes sense?

dgalaktionov:
I'd be happy with that, yet i'm unsure how consistent would that be with other languages: my layout would be the Russian analogous for the English `qwerty_with_symbols`. However, that does not seem to be the default layout for English.

Will await for a final decision.

menny:
If I understand correctly, you took the original [russian_qwerty](https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/e542d2349ef797e9f014e1464fdb442a6eb0ff26/addons/languages/russian2/pack/src/main/res/xml/russian_qwerty.xml) and added a few popup characters. You have not changed the layout. Correct?

dgalaktionov:
Yes, that's right!

menny:
great. In that case, please revise your PR to modify `russian_qwerty` instead of creating a new layout.

dgalaktionov:
Done! Replaced the default keyboard with mine, and updated the PR description to be a more accurate summary of what has been done.

menny (approved):

PR #3684, opened 2023-04-20
New layout with wider spacebar

joelsgp:
image

I prefer the wider spacebar on the right because I don't have to reach as far with my thumb so it gets strained less. Also, it's closer to what GBoard has so maybe easier for new users to get used to. We could maybe add one that stretches the spacebar to the left instead for left-handed people.

I'm not experienced with android studio or this project so I still need help 1. setting the `id` field for the new entry in english_keyboards.xml 2. changing the colour of the new bottom row - they're the same look and feel as the regular letter keys right now which I don't think is right

many thanks!


menny:
this should be a "Generic bottom row", not a new English layout. Can you convert this layout to a generic bottom row?

joelsgp:
Ok, well I figured out where bottom row options are in the settings and it turns out that's just what I was looking for so I'll close this. A way to search the settings menu or some improved help would be useful but that's a separate issue. thanks

Issue #3230, opened 2022-02-01
Speech to text not working, needs com.google.android.voicesearch


backupandforkacc:
### Steps to reproduce
1. Click on microphone symbol in keyboard to open STT

### Actual behaviour

**Android OS version:** LOS 18.1-2022-01-22-microG-walleye

**Device manufacturer and model:** Pixel 2

**List of installed add-ons (like languages, or themes):** Konele, google app, Konele-service

Info:
Speech to text generally works finde, p.ex. in Duolingo. 
I will try to find out how ASK calls the STT service. As the app is looking for com.google.android.voicesearch, this might be not a proper implementation. Something [like implemented here](https://github.com/Kaljurand/K6nele/blob/master/app/src/main/java/ee/ioc/phon/android/speak/activity/RecServiceSelectorActivity.java) might be helpful.

adeyo50:
I'm a similar problem as well. (android 10)... I have my default assistant app set as k6nele, my voice input and recognition services both set as kaldi/vosk recognizer... Yet somehow when I click the mic button it brings up Google speech to text. Shouldn't it use our defaults and NOT pull up the com.google.android.voicesearch, as @backupandforkacc pointed out? If there are no plans to do so, I'll have to jump to something else.... privacy has become too important to me =(

benmordecai:
I would also like a clear and straightforward way to deploy a vosk-based STT service from Anysoft.

anulin:
I also have this issue for my Volla X23 and Vollaboard voice input

noraj:
Sayboard does that

Sayboard (An open-source on-device voice IME (keyboard) for Android using the Vosk library)
https://f-droid.org/packages/com.elishaazaria.sayboard/

That force you to change of keyboard, but you can configure it to automatically listen right after switching and switch back to the previous keyboard (Anysoftkey) right after ending the recording.

That could be a workaround while Anysoftkeyboard makes the SST provider configurable.

ferromint:
> Sayboard (An open-source on-device voice IME (keyboard) for Android using the Vosk library)

On my Google-free LineageOS (so no proprietary Google STT available) Sayboard was also my first STT app which is working as @noraj described (provides quite good results meeting my expectations). Very recently I stumbled across [FUTO Voice Input](https://voiceinput.futo.org) which recognizes the spoken text with an even higher accuracy.

[...]

Issue #3004, opened 2021-09-23
microphone/ voice TTS uses google NOT default


adeyo50:
### Steps to reproduce
1. Tap microphone button on keyboard
2. Google TTS engine is used
3.

### Expected behaviour
 flite TTS (my default) should be used
### Actual behaviour

**Android OS version:**
10
**Device manufacturer and model:**
SM N986U
**List of installed add-ons (like languages, or themes):**
None

adeyo50:
Honestly, I'm no developer but this is the only way I could figure out how to provide feedback. Also, I have properly set my default tts engine and tested it. Unfortunately, I can't seem to override Google's TTS using anysoft canary. I use anysoft for privacy and this will force me to alternatives bc I use the microphone a lot. Can somebody take this on? Thanks

adeyo50:
Oh, and it's a non rooted phone

PR #3689, opened 2023-04-25
switch order of search for speech-to-text

drew-sinha:
This is a modification to make Google stt the non-default for ASK stt. This fix simply reverses the order of searches for the Google stt app and the intent-api based pm query. Directly addresses #3230, and indirectly addresses #3004 - specifically in the context of using vosk-android-service (as this stt service is at least partly functional as well; also see https://github.com/alphacep/vosk-android-service/issues/33). Could consider additional user preference to control this ordering in the future. 

Change not rebuilt/tested for "stability" given minimal change to underlying codebase.

Edit: included note about #3230


menny (approved):

menny:
Thank you for this important addition.

I also want to note that if no voice service was detected, we direct the user to download it from the Play Store. I assume vosk is not available in the Play Store?

PR #3690, opened 2023-04-25
switch order of search for speech-to-text (fixed formatting)

drew-sinha:
Sigh... redo of the prior PR that @menny had commented on (my newb self rebased and messed up the repo history; this should hopefully be a stable fix that passes the formatting tests).

This feature edit switches the order of searches for available stt services to prioritize non-google based applications (e.g. vosk android service). This fix directly addresses ASK issue #3230 (see also vosk-android-service https://github.com/alphacep/vosk-android-service/issues/33) and indirectly addresses #3004. Can consider adding user-based preferences on whether to prioritize a given service in the future.

Changes not tested for stability given minimal change to codebase.

To address @menny's question on the original PR - unfortunately, vosk-android-service is not available on the play store yet.

PR #3695, opened 2023-05-02
Fix english duplicates

drew-sinha:
This fix addresses a build-time error for the recent build [498dab3] in which generating an english word list fails on Windows due to inappropriate heuristic matching for the word list build directory (relies on searching for '/build/' in the folder path to avoid checking itself against duplicate words; added another condition for the alternate filesep).


drew-sinha:
Oops. Contaminated this branch with the other PR from #3690

PR #3697, opened 2023-05-02
Fix build folder search heuristic for duplicate word finding

drew-sinha:
This fix addresses a build-time error for the recent build [https://github.com/AnySoftKeyboard/AnySoftKeyboard/commit/498dab3a1cba1b197b7b48aea40e7e595c50d299] in which generating an english word list fails on Windows due to inappropriate heuristic matching for the word list build directory (relies on searching for '/build/' in the folder path to avoid checking itself against duplicate words; added another condition for the alternate filesep).

(sorry... here's try 3.)


menny:
@drew-sinha, thanks for figuring this out. I believe using [File.separator](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#separator) would make more sense here.

drew-sinha:
Done. Looks a little ugly, but works.

menny (approved):

PR #3737, opened 2023-06-02
Don't show voice result dialog when only 1 result

abb128:
Currently, AnySoftKeyboard shows a voice recognition result dialog even if there's only one result. I assume this dialog is intended to allow picking between multiple results, but when there's only one result, it just adds an unnecessary step. This change shows the dialog only when there's two or more results, and skips to typing the text instantly if there's only one.

Example of the dialog with only one result:
image

### DIFF
--- a/ime/voiceime/src/main/java/com/google/android/voiceime/ActivityHelper.java
+++ b/ime/voiceime/src/main/java/com/google/android/voiceime/ActivityHelper.java
@@ -64,9 +64,9 @@ public class ActivityHelper extends Activity {
         && data.hasExtra(RecognizerIntent.EXTRA_RESULTS)) {
       ArrayList<String> results = data.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS);
 
-      if(results.size() > 1) {
+      if (results.size() > 1) {
         createResultDialog(results.toArray(new String[results.size()])).show();
-      } else if(results.size() > 0) {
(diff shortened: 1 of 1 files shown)


ime/voiceime/src/main/java/com/google/android/voiceime/ActivityHelper.java
```diff
@@ -63,7 +63,14 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         && data != null
         && data.hasExtra(RecognizerIntent.EXTRA_RESULTS)) {
       ArrayList<String> results = data.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS);
-      createResultDialog(results.toArray(new String[results.size()])).show();
+
+      if(results.size() > 1) {
+        createResultDialog(results.toArray(new String[results.size()])).show();
+      } else if(results.size() > 0) {
```
menny, line 69: For readability, I would change this to `== 1`
```suggestion
      } else if(results.size() == 1) {
```

menny (changes requested):

menny:
thanks for this

ime/voiceime/src/main/java/com/google/android/voiceime/ActivityHelper.java
abb128: Updated

menny (approved):

PR #3744, opened 2023-06-19
Add ǐ and ǎ to qwerty, to complete pinyin

nstbayless:
Added ǎ and ǐ, which more or less completes the ability to type [pinyin](https://en.wikipedia.org/wiki/Pinyin) in the QWERTY keyboard (not counting ü-tones). The vowels+tones used in pinyin are:

```
āáǎàa
ēéěèe
īíǐìi
ōóǒòo
ūúǔùu
ǖǘǚǜü
```

(However, the `ü` can also be written as a `u` (with no umlaut) or v or even yu depending on context and style.)

Before this pull request, the ability to type in pinyin is nearly complete on AnySoft keyboard in QWERTY except for just two tones that are missing.  It seems like an important and low-hanging fruit to round off the ability to write in pinyin. Chinese is the second most widely spoken language in the world, after all.

### DIFF
--- a/addons/languages/english/pack/src/main/res/xml/qwerty.xml
+++ b/addons/languages/english/pack/src/main/res/xml/qwerty.xml
@@ -26,13 +26,13 @@
         <Key android:codes="116" android:popupCharacters="5țť\u0163τ"/>
         <Key android:codes="121" android:popupCharacters="6ýÿψ"/>
         <Key android:codes="117" android:popupKeyboard="@xml/popup_qwerty_u"/>
-        <Key android:codes="105" android:popupCharacters="8ìíîïłīι*"/>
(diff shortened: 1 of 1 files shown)


menny:
The change does include `ǎ` not `ǔ`.
Is that intentional?

nstbayless:
> The change does include `ǎ` not `ǔ`. Is that intentional?

`ǔ` is already on the keyboard.

menny:
okay. can you change the title to match the change?

nstbayless:
> okay. can you change the title to match the change?

My bad. Fixed.

menny (approved):

menny:
thank you!

PR #3779, opened 2023-08-01
Fix hardcoded google voice input check

abb128:
Currently, ImeTrigger is hardcoded to only look for voice IME subtypes whose package names are prefixed with `com.google.android`. This means that only Google Voice Input is allowed to be picked by ImeTrigger.

From my understanding, it's adequate to check that the subtype equals `"voice"`. Any input method subtype advertising its mode as `"voice"` is probably capable of doing voice input. So I'm not certain why this check was there in the first place, other than to prioritize Google's Voice Input on the belief that Google's probably works better than whatever else is installed by the user or vendor.

This PR removes this check to allow non-Google voice inputs to be picked by ImeTrigger. Previously it was still possible for third party voice inputs to be picked by IntentApiTrigger instead, but the intent API isn't as well-integrated as an input method and it's a better experience to use an IME instead, so this PR also prioritizes ImeTrigger over IntentApiTrigger.

### DIFF
--- a/ime/voiceime/build.gradle
+++ b/ime/voiceime/build.gradle
@@ -5,6 +5,8 @@ apply from: "${rootDir}/gradle/android_general.gradle"
 dependencies {
 
     implementation project(path: ':ime:base')
+
+    testImplementation project(path: ':ime:base-test')
 }
 
 android {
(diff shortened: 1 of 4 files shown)


menny (commented):
This is great!
Can you add tests for this?

abb128:
@menny I attempted to write some tests. Let me know how it looks

menny (approved):

menny:
thank you!

PR #3802, opened 2023-10-01
Temporary fix for Yochees Light theme having unreadable icons

### LINKED ISSUE #2439: Keyboard theming issue when Quick-text editing top row is enabled
### Steps to reproduce
1. Set theme to yochees light 

### Expected behaviour
Top row should be visible

### Actual behaviour
Its not visible clearly because the text colour also is white.

**Android OS version:**
Android Q

Xiaomi redmi note 7
None
Screenshot_20200822-150722_KISS_launcher

Note : Almost similar issue in 
Plain Light theme.

LunarEclipse363:
This isn't the "correct" way to fix this, which would be creating a separate set of icons for the yochees light theme, but it fixes #2439 and since no-one touched that bug in 3 years, I think this is good enough.

image

### DIFF
--- a/ime/app/src/main/res/values/styles_yochees_keyboard_theme.xml
+++ b/ime/app/src/main/res/values/styles_yochees_keyboard_theme.xml
@@ -183,7 +183,8 @@
 
     <!-- light theme -->
     <style name="YocheesLightKeyIconTheme"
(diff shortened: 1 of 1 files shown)


menny (approved):

menny:
Yep, makes sense as a quick usability fix. Thanks

PR #3863, opened 2024-01-02
en dvorak: add ð & þ

toastal:
Eth & thorn, ð & þ, were historic letters to English & still common in Icelandic words. There seems to be recurring movements online to use them in English today too to supplant the most common digraph, “th”, for [θ] [ð] sounds.

Placement was at the beginning of the list as these characters have higher codepoint values than the others (altho the order doesn’t have an order I personally understand—and ideally users could reorder them).


menny (changes requested):
please move these characters to the end of the popup list

menny (approved):
Thank you

PR #3903, opened 2024-02-07
Add Gershayim to Hebrew

paulfertser:
This symbol is used extremely often in many abbreviations (e.g. words for kilogram, doctor etc.) and typing Geresh twice is not a proper substitute.
The change wasn't tested as I have no dev environment set up.


menny (approved):

menny:
thank you.

paulfertser:
Hi, thank you for this software and for maintaining it!
I wasn't able to find an apk in the Github CI artifacts, am I looking at the wrong place? How else can I possibly get an updated version given that F-droid language pack seems to get no updates in years?

PR #4008, opened 2024-06-12
Update packs_md_update.yml

Bangyourkitty:


menny:
We don't need this, we have a docker image with the required java version

PR #4035, opened 2024-07-28
Update russian_qwerty.xml

Azim:
Revert char order change introduced in 86be98b

this is just driving me nuts, why would you suddenly change it like that. (ik that was an old commit but i only noticed and updated language pack now, and there isnt a way to revert it now)

Honestly, much better solution would be to make it into a separate option, but i couldnt find an easy way of doing that


menny (approved):

menny:
You'll need to rebase this change and force-push

Azim:
i might've did it incorrectly
lmk if i need to redo it

Azim:
merge?

Azim:
Now that its merged, how do i download it? seems to be gone from play store

PR #4041, opened 2024-08-28
added Polytonic Greek Keyboard

semaurer01:
I wanted a better way to type in Polytonic Greek. [Apptide](https://ancient-languages-pack-for-anysoftkeyboard.en.aptoide.com/app) has an ancient languages pack that supports Polytonic Greek, but it is hard to use because of the large popup menus. So I have created one that uses combining diacritics. and lists them in a row for easy access. now there is some difficulty because a lot of systems including mine(android 15) do fine if one types: Vowel, Breathing Mark, Tonos(ἔ); but have trouble if one types: Vowel, Tonos Breathing Mark, (έ̓). but I know not how to solve this problem in AnySoftKeyboard. I don't think I did anything "Copyright-able" so do with this keyboard as you please. I do not know whether it really belongs in the Greek package or in a new package for itself.

### DIFF
--- a/addons/languages/greek/pack/src/main/res/xml/greek_qwerty.xml
+++ b/addons/languages/greek/pack/src/main/res/xml/greek_qwerty.xml
@@ -3,15 +3,15 @@
 <Keyboard xmlns:android="http://schemas.android.com/apk/res/android"
     android:keyWidth="10%p"><!-- "android:keyWidth" specify the default width of a key. In this example, 10% of parent (the whole keyboard) -->
 
-    <!-- Key attributes:
-     "android:codes" : a comma separated unicode values of the keys. If you specify more than one code, then the other codes are accessible via multi-tap.
-     "android:popupCharacters" : characters to show on long-press popup keyboard
-     "android:keyLabel" : the text to show on the key. If this attribute is missing, the first code in "android:codes" will be used.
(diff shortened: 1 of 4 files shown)


addons/languages/greek/pack/src/main/res/xml/greek_qwerty.xml
menny, line 1: I think you do not have changes in this file, right?
Please revert the changes, as this will not pass CI checks.

menny (changes requested):
thanks for the new layout. I have one request for a change

addons/languages/greek/pack/src/main/res/xml/greek_qwerty.xml
semaurer01: I do not understand the objection. Being new to git I know not: what are CI checks? nor do I know how to revert changes to a specific file. please explain the issue in more detail.

menny: You have introduced changes in this file. These changes are not required and, in fact, include code-style changes that are not compatible with the project.

semaurer01: I think I have resolved this issue.

menny (approved):

PR #4091, opened 2024-12-01
Enabled input methods fix

mirfatif:
Fix Settings.Secure.ENABLED_INPUT_METHODS not readable on SDK 34+

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/ui/settings/setup/SetupSupport.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/ui/settings/setup/SetupSupport.java
@@ -31,11 +31,7 @@ public class SetupSupport {
     if (TextUtils.isEmpty(defaultIME)) return false;
 
     ComponentName defaultInputMethod = ComponentName.unflattenFromString(defaultIME);
-    if (defaultInputMethod.getPackageName().equals(myPackageName)) {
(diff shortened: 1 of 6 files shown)


menny:
interesting. How did you get this error?

mirfatif:
I switched my custom ROM to Android 15, rebuilt ASK from upstream, installed the APK, and it crashed on first launch.

menny:
I see. Can you take this on?

Also, this should not happen, as we target API level 34: https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/main/gradle/root_all_projects_ext.gradle#L8.
Regardless, it still needs to be tackled.

mirfatif:
By `SDK 34+` I meant SDK 34 and above :)

From [docs](https://developer.android.com/reference/android/provider/Settings.Secure#ENABLED_INPUT_METHODS):

> Note: This setting is not readable to the app targeting API level 34 or higher. use [InputMethodManager.getEnabledInputMethodList()](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#getEnabledInputMethodList()) instead.

So this should happen on API level 34 too.

menny:
I wonder why it does not crash for me. In this case, it should be fixed.

menny (approved):

PR #4242, opened 2025-06-16
Update cyrillic_qwerty.xml

bkomdoshev:
Added "ј" for long pressed д

### DIFF
--- a/addons/languages/russian2/pack/src/main/res/xml/cyrillic_qwerty.xml
+++ b/addons/languages/russian2/pack/src/main/res/xml/cyrillic_qwerty.xml
@@ -30,7 +30,7 @@
         <Key android:codes="1088" android:keyLabel="р"/>
         <Key android:codes="1086" android:keyLabel="о" android:popupCharacters="ӧөӫѳѻ"/>
         <Key android:codes="1083" android:keyLabel="л" android:popupCharacters="љѧѩ"/>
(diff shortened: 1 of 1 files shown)


menny:
Interesting. Why specifically `j`? And there are no other latin letters that need to be added?

bkomdoshev:
Excuse me.

I found the missing characters on the 'Е' and 'е' keys. The 'Ј' and 'ј' characters would be more correctly placed together with 'Д' and 'д'.
I not found Д in xml file and cant add Ј for Д.

bkomdoshev:
> Interesting. Why specifically `j`? And there are no other latin letters that need to be added?

It is not latin J and j, it is Cyrillic letters https://en.wikipedia.org/wiki/Je_(Cyrillic).

Now I suggest moving the characters Ј and ј to the letters Д and д. 

In older versions of Android, [this](https://androspace.ru/1557436-altajskaja-klaviatura.html) app worked with ASK, but now it doesn’t install on Android.

menny:
@bkomdoshev , can you rebase this branch and push? It seems that your base commit has issues.

bkomdoshev:
> @bkomdoshev , can you rebase this branch and push? It seems that your base commit has issues.

Can your add the Ј and ј letters for Д and д keys?

I'm just delete this pool request.

PR #4250, opened 2025-06-21
Adding missing popup character for Cyrillic

menny:
Fixes https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4242


menny:
@bkomdoshev, is this right?

bkomdoshev:
> @bkomdoshev, is this right?

Hi! Yes! But i do not know how it work to big letters(for example: Д->J), if it work- it will be right. 
Big Thanks for you!

bkomdoshev:
Both keys works fine! 
Many thanks for you!
Best regards!

PR #4285, opened 2025-07-06
Improved Nias wiki keyboard layouts

sslaia:
Improved the nias wiki layout:
- removed the double click feature for producing apostrophe sign in the Nias language
- added the necessary ö and ŵ keys (no need of double clicking)
- nias wiki layout is now enabled by default
- updated the graphics to reflect the new keyboard layouts
- updated the release notes


addons/languages/nias/pack/src/main/res/xml/nias_keyboards.xml
```diff
@@ -11,7 +11,7 @@
         nameResId="@string/nias_qwerty" />
     <Keyboard
         defaultDictionaryLocale="nia"
-        defaultEnabled="false"
+        defaultEnabled="true"
```
menny, line 14: Usually, only one layout will be default. If you have two, both will be enabled to new users which may give a weird user experience.

addons/languages/nias/pack/src/main/res/xml/nias_autotext.xml
```diff
@@ -20,13 +20,13 @@
     <!-- this dictionary is used for common typing mistakes -->
     <!-- the attribute 'src' will be replace by the node's value: -->
     <!-- rght will be converted to right -->
-    <word src="yahowu">ya'ahowu</word>
+    <!-- <word src="yahowu">ya'ahowu</word>
```
menny, line 23: you don't want all these auto-fixes?

addons/languages/nias/apk/src/main/play/release-notes/en-US/alpha.txt
```diff
@@ -1 +1,12 @@
-* New owner
+1. Improved the nias wiki layout:
```
menny, line 1: The file can not be more than 500 characters long (this is a limitation from Google Play).
From the error message: `whatsnew file can not be longer than 500 characters! Currently 598`

sslaia: Sorry, I didn't know about this. I'll update the relevant files and submit the pull request later.

PR #4293, opened 2025-07-09
chore: make cyrillic keyboard reflect slavic etymology and common typing standards

gordonel:
This keyboard layout is primarily used to type etymological interslavic: https://en.wikipedia.org/wiki/Interslavic

Some of the pop-up character assignments wouldn't really make sense to interslavic users, as well as those studying Slavic languages or typing Old Church Slavonic, because the pop-up characters don't have anything to do with the main characters assigned to those keys. This PR fixes this issue.

<h3>Summary of the changes</h3>
- "џ" is not a variant of "ц", but it's a voiced equivalent of the unvoiced "ч" in Interslavic and Serbo-Croatian
- "ѫ" (and it's iotated ligature ѭ) are not used in modern languages because the sounds they used to make changed to "y" sounds
- same thing with ѧ, ѩ. They have nothing to do with л, but in the majority of languages the sounds they used to make migrated to "e" sounds.
- ѥ is an iotated ligature derivative of є
- ѕ is assigned sort-of correctly, but it would be more convenient for Serbo-Croatian speakers to have it also at я position because that's where it is in their [keyboard layout](https://en.wikipedia.org/wiki/Serbian_Cyrillic_alphabet), so I duplicated it.

<h3>Future Considerations</h3>
- Remove j assignment from д, because in Slavic languages it makes the English "y" sound as in "Yes"

I'm working on a dedicated interslavic language add-on which will be based on Russian2 with these assignments, but this is a good temp fix imo

### DIFF
--- a/addons/languages/russian2/pack/src/main/res/xml/cyrillic_qwerty.xml
+++ b/addons/languages/russian2/pack/src/main/res/xml/cyrillic_qwerty.xml
@@ -6,10 +6,10 @@
     <Row android:keyWidth="8.33%p">
         <!-- йцукенгшщзхъ    -->
         <Key android:codes="1081" android:keyLabel="й" android:popupCharacters="1їӣӥ" android:keyEdgeFlags="left"/>
-        <Key android:codes="1094" android:keyLabel="ц" android:popupCharacters="2ћџҵ"/>
-        <Key android:codes="1091" android:keyLabel="у" android:popupCharacters="3ўүұӯӱӳѹ"/>
+        <Key android:codes="1094" android:keyLabel="ц" android:popupCharacters="2ћҵ"/>
+        <Key android:codes="1091" android:keyLabel="у" android:popupCharacters="3ўүұӯӱӳѹѫѭ"/>
         <Key android:codes="1082" android:keyLabel="к" android:popupCharacters="4ќқҝҟҡӄ"/>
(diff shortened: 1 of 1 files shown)


menny:
thanks for this thoughtful change. As I am not an expert of this, I have a question:
- this pack is for Russian support. How does this layout (cyrillic) stands with the `russian` layout?
- I see that your reasoning is that cyrillic layout should reflect slavic standards. Why is that?
- Can you update the change log for this? https://github.com/AnySoftKeyboard/AnySoftKeyboard/tree/main/addons/languages/russian2/apk/src/main/play/release-notes

fedigord:
> - this pack is for Russian support. How does this layout (cyrillic) stands with the `russian` layout?

Not sure about what you mean by this question exactly, but the ability to type Russian with this layout is unchanged. The popup characters are not used in modern Russian, but they can be used to write early modern Russian when it had a different orthography.

> - I see that your reasoning is that cyrillic layout should reflect slavic standards. Why is that?

I think you generally want pop-up characters to have some relation to the primary characters in their keys, wouldn't you?

> - Can you update the change log for this? https://github.com/AnySoftKeyboard/AnySoftKeyboard/tree/main/addons/languages/russian2/apk/src/main/play/release-notes

Will do later

menny:
@gordonel I want to get this merged, if anything is unclear in my comments, lmk.

PR #4382, opened 2025-08-19
Added Norwegian (Nynorsk) keyboards

EiriKGaratun:
Modified the Norwegian keyboards to use a Nynorsk dictionary. Also slightly modded the Dvorak layout. The dictionary was only a alphabetical list of allowed Nynorsk words, so the suggestions aren't sorted by most likely.


menny:
This is a completely new pack. Was that your intention?
The other option would be to add these changes to the addons/languages/norwegian pack

EiriKGaratun:
Nynorsk is a distinct variety of Norwegian, with unique spelling and slightly different grammar. It has its own locale and ISO 639 code: nn. It therefore requires a keyboard with a distinct dictionary, which is why I created it as a new pack. It seemed natural to keep the varieties distinct, as only a fraction of Norwegian speakers use Nynorsk. Allowing only those who use it to download it. But I could see an argument for keeping having them as a single pack. But its quite commonplace to treat them as seperate languages on tech platforms e.g. Wikipedia, Linux.

menny (approved):

menny:
Thank you for your contribution
I will get this published later this week

Issue #4341, opened 2025-08-01
Long-tapping emoji key leads to crash


Flashwalker:
Hello.

If i make a clean install v.1.12.7138 and then I restore my custom settings from a backup, this issue appears:

long-tapping emoji key leads to app crash.

[AnySoftKeyboardPrefs.zip](https://github.com/user-attachments/files/21541247/AnySoftKeyboardPrefs.zip)

lkelemen:
on which version of android?

systemoperator:
@Flashwalker Only, when long-tapping it, or whenever you press the emoji key? I've opened an issue for pressing the emoji key a few months ago: #4195

Flashwalker:
> [@Flashwalker](https://github.com/Flashwalker) Only, when long-tapping it, or whenever you press the emoji key? I've opened an issue for pressing the emoji key a few months ago: [#4195](https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/4195)

Only, when long-tapping it, and only with my settings from a backup.

slecache:
Same here. The keyboard doesn't respond and then crashes after a long press on emoji key or after a clic on emoji menu in the app parameters.
It occures since the last update (from f-droid).

di72nn:
Probably a duplicate of #4300. Try this: https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/4300#issuecomment-3119597357

nickleverton:
I am also experiencing this hang followed by a crash after long-press on emoji button.  I have checked the suggestion re #4300 and I already have "settings_key_default_emoji_gender" set to "person", but the hang and subsequent crash are happening regardless.

I have managed to get two logcat dumps out, I don't know if they may be helpful ?  I'll attempt to upload them in a moment.

EDIT: it's quite an old device, Android 9 is the latest available for it.

****** Application name: AnySoftKeyboard (com.menny.android.anysoftkeyboard) v1.12.7138 release 7138. Installed on 1754124192316, first release installed was 5781.
****** Device information:
BRAND:Sony
DEVICE:G8441
Build ID:47.2.A.11.228
changelist number:47.2.A.11.228
MODEL:G8441
PRODUCT:G8441
TAGS:release-keys
VERSION.INCREMENTAL:3311891731
VERSION.RELEASE:9
VERSION.SDK_INT:28
Locale:en_GB
[...]

Issue #4300, opened 2025-07-15
Opening Emojis freezes the Keyboard


zyweii-3:
I've had the issue for a while where once every few times, trying to open the Emoji window Freezes the keyboard, and i'm unable to close it or type anything. Recently, it started happening every time i tried acessing the emojis.

Staying on the subject of emojis, a few have been added to androids that are not available in the ASK's emoji keyboard.

Mersadon:
Yes. This is happening to me as well after the last update.
Also, I noticed that I cannot even access emojis in the AnySoftKeyboard app, because it crashes each time when I click on emoji settings in the bottom right corner.

NicoHood:
Related: https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/1860

I also noticed that this happens when opening the emoji settings inside the settings app as i tried to disable them as a workaround. 100%reproducible every time!

zyweii-3:
Yes it did made my app crash when trying to change the setting in app for me too.

What I ended up doing after a while is that I uninstalled and installed back ASK. A pain, but now, I can re-use emoticons again without it crashing again!

Also, the error log I was getting was something along the lines of: "default skin tone can't be set as [none]" or something like that.

Mersadon:
OK. I confirm that reinstalling the app helped.
But you must not backup your settings before the reinstallation and restore them after the reinstall, because then it happens again.

di72nn:
I don't really have time to figure out how it is supposed to work, but the work around for now is the following:

1. Back up app preferences.
2. Set `settings_key_default_emoji_gender` to `person` in the backup file (you don't need to add it, **it should already exist in the file**). Update: removing this `pref` altogether should probably have the same effect, but I haven't tested it.
3. Import the modified backup.

Steps to reproduce with the latest version:
1. Clean-install app.
2. Back up app preferences.
3. Set `settings_key_default_emoji_gender` to any non-default value (e.g., `man`) in the backup.
4. Import modified preferences backup. Like this one: [AnySoftKeyboardPrefs_broken.xml.txt](https://github.com/user-attachments/files/21435163/AnySoftKeyboardPrefs_broken.xml.txt)
5. Open the main app screen and press the emoji tab at the bottom - freeze and crash.

Flashwalker:
> * Back up app preferences.
> * Set `settings_key_default_emoji_gender` to `person` in the backup file.
> * Import the modified backup

[...]

Issue #4195, opened 2025-05-22
Since update keyboard crashes instantly, whenever opening the emoji selector


systemoperator:
Whenever I want to select an emoji, the keyboard crashes instantly.

It started with the AnySoftKeyboard Canary update just before v1.12.913 release 8541.
With AnySoftKeyboard Canary v1.12.913 release 8541 the error still persists.

Here are the details:

Hi. It seems that we have crashed.... Here are some details:
****** UTC Time: 13:22:52 22.05.2025
****** Application name: AnySoftKeyboard Canary (com.menny.android.anysoftkeyboard) v1.12.913 release 8541. Installed on 1747867384784, first release installed was 7137.
******************************
java.lang.AssertionError
****** Exception message: impossible
----- FULL REPORT -----
****** Trace trace:
at java.lang.Enum$1.create(Enum.java:272)
at java.lang.Enum$1.create(Enum.java:262)
at libcore.util.BasicLruCache.get(BasicLruCache.java:58)
at java.lang.Enum.getSharedConstants(Enum.java:289)
at java.lang.Enum.valueOf(Enum.java:244)
at com.anysoftkeyboard.keyboards.AnyKeyboard$AnyKey.stringsToEnum(SourceFile:29)
at com.anysoftkeyboard.keyboards.AnyKeyboard$AnyKey.<init>(SourceFile:73)
at com.anysoftkeyboard.keyboards.AnyKeyboard.createKeyFromXml(SourceFile:12)
at com.anysoftkeyboard.keyboards.AnyPopupKeyboard.createKeyFromXml(SourceFile:1)
at com.anysoftkeyboard.keyboards.Keyboard.loadKeyboard(SourceFile:184)
at com.anysoftkeyboard.keyboards.AnyKeyboard.loadKeyboard(SourceFile:7)
at com.anysoftkeyboard.keyboards.AnyKeyboard.loadKeyboard(SourceFile:6)
at com.anysoftkeyboard.keyboards.AnyPopupKeyboard.<init>(SourceFile:7)
at com.anysoftkeyboard.quicktextkeys.ui.QuickKeysKeyboardPagerAdapter.instantiateItem(SourceFile:132)
at androidx.viewpager.widget.ViewPager.addNewItem(SourceFile:10)
at androidx.viewpager.widget.ViewPager.populate(SourceFile:43)
at androidx.viewpager.widget.ViewPager.populate(SourceFile:1)
[...]

PR #4407, opened 2025-08-23
Fix long-press emoji crash

menny:
The application was crashing when opening the emoji keyboard due to ProGuard/R8 stripping the `values()` and `valueOf()` methods from Enum classes. This was happening because these methods were being used via reflection, and ProGuard/R8 could not detect this usage.

I have added a ProGuard rule to keep these methods for all Enum classes, which resolves the crash.

Signed-off-by: Gemini-Code-Assistant

Fixes: https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/4341
Fixes: https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/4195
Fixes: https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/4300

### DIFF
--- a/ime/app/proguard-rules.txt
+++ b/ime/app/proguard-rules.txt
@@ -36,3 +36,8 @@
 -keep public class * extends android.content.BroadcastReceiver
 
 -keep public class * extends java.lang.Exception
+
+-keepclassmembers enum * {
+    public static **[] values();
+    public static ** valueOf(java.lang.String);
+}


di72nn:
This change does seem to resolve the `AssertionError: impossible` issue, but not #4300 - that one needs some actual changes is logic. It can be easily reproduced as described [here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/issues/4300#issuecomment-3119597357).

NicoHood:
Its also still broken for me

PR #4541, opened 2025-11-29
Better gesture typing accept words whose starting char is close to first gesture path corner

doodhout:
I tried improving the gesture typing mechanics by not requiring the user to start their swiped word with the correct first character.

Example:
- Before this change: you want to swipe the word Juggernaut, but you start the gesture on the key next to it. Result: you will never get the word you wanted, because AnySoftKeyboard dismisses all words not starting with the first letter of the gesture path.
- After this change: AnySoftKeyboard now also suggests words whose path aligns with the gesture path without requiring the first letter to be correct, so when you swipe the gesture path for Juggernaut but you start on the letter K, now you still get the suggestion for the word Juggernaut, since the gesture path in total aligns best with that word.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
@@ -26,6 +26,19 @@ public class GestureTypingDetector {
   private static final int CURVATURE_NEIGHBORHOOD = 1;
   private static final double MINIMUM_DISTANCE_FILTER = 1000000;
 
+  /**
+   * Maximum squared distance from gesture start point to accept a word's starting key.
+   * This allows for imprecise gesture starts while filtering obviously wrong candidates.
+   * Value is approximately 1.5 key widths squared (assuming ~100 pixel keys).
+   */
+  private static final int START_KEY_PROXIMITY_THRESHOLD_SQUARED = 22500;
+
+  /**
+   * Penalty factor for words that start near but not on the exact starting key.
+   * Lower value = less penalty, higher value = more penalty.
+   */
+  private static final double PROXIMITY_PENALTY_FACTOR = 0.1;
+
   // How far away do two points of the gesture have to be (distance squared)?
   private final int mMinPointDistanceSquared;
 
@@ -329,10 +342,17 @@ public class GestureTypingDetector {
       final char[][] words = mWords.get(dictIndex);
       final int[] wordFrequencies = mWordFrequencies.get(dictIndex);
       for (int i = 0; i < words.length; i++) {
-        // Check if current word would start with the same key
+        // Check if current word starts with a key close to the gesture start point
         final Keyboard.Key wordStartKey = mKeysByCharacter.get(Dictionary.toLowerCase(words[i][0]));
-        // filtering all words that do not start with the initial pressed key
-        if (wordStartKey != startKey) {
+        if (wordStartKey == null) {
+          continue; // Character not found on keyboard
+        }
+
+        // Calculate squared distance from gesture start to word's starting key
+        final int distanceSquared = wordStartKey.squaredDistanceFrom(corners[0], corners[1]);
+
+        // Filter out words whose starting key is too far from gesture start
+        if (distanceSquared > START_KEY_PROXIMITY_THRESHOLD_SQUARED) {
           continue;
         }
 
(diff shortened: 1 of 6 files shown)


ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -26,6 +26,19 @@ public class GestureTypingDetector {
   private static final int CURVATURE_NEIGHBORHOOD = 1;
   private static final double MINIMUM_DISTANCE_FILTER = 1000000;
 
+  /**
+   * Maximum squared distance from gesture start point to accept a word's starting key.
+   * This allows for imprecise gesture starts while filtering obviously wrong candidates.
+   * Value is approximately 1.5 key widths squared (assuming ~100 pixel keys).
+   */
+  private static final int START_KEY_PROXIMITY_THRESHOLD_SQUARED = 22500;
```
menny: This should be a resource. Probably add that to ime/app/src/main/res/values/dimens.xml. For example, instead of 100 pixels, you may add `<dimen name="gesture_typing_proximity_threshold">20dp</dimen>` (or whatever value works for you). Then you can use `context.getResources().getDimensionPixelSize(R.dimen.gesture_typing_proximity_threshold)`

Or something like that.

menny (changes requested):
Thank you so much! This is very helpful. Can you attend the comments?

ime/app/src/main/java/com/anysoftkeyboard/keyboards/Keyboard.java
```diff
@@ -612,9 +612,9 @@ public boolean isInside(int x, int y) {
       final boolean topEdge = (edgeFlags & EDGE_TOP) != 0;
       final boolean bottomEdge = (edgeFlags & EDGE_BOTTOM) != 0;
       return (x >= this.x || (leftEdge && x <= this.x + this.width))
-          && (x < this.x + this.width || (rightEdge && x >= this.x))
+          && (x <= this.x + this.width || (rightEdge && x >= this.x))
```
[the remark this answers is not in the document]
doodhout: Good catch, I realise now that I tried fixing the 1 pixel horizontal gap from the test case in completely the wrong way.

Original reasoning: one of the new unit-tests picks a spot in the middle of two keys, but of course that spot falls into the gap between the keys, causing the GestureTypeDetector's check `if (k.isInside(corners[0], corners[1])) {` to not match any key, thus `startKey` is null, thus it will return zero candidates.

I will have to revert this change and rethink how to handle that situation, because if it's acceptable to not start on the correct key, then it should also be acceptable to not start on any key.

doodhout:
I'm working through your suggestions.
I need to fix some existing unit-test and then need to add these requested unit-tests.
Thanks for your patience. :-)

doodhout:
@menny I have addressed all comments and suggestions, this should be ready for another review.

Thanks!

ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -318,21 +333,23 @@ public ArrayList<String> getCandidates() {
...
+        // Check if current word starts with a key close to the gesture start point
         final Keyboard.Key wordStartKey = mKeysByCharacter.get(Dictionary.toLowerCase(words[i][0]));
-        // filtering all words that do not start with the initial pressed key
-        if (wordStartKey != startKey) {
+        if (wordStartKey == null) {
+          continue; // Character not found on keyboard
+        }
+
+        // Calculate squared distance from gesture start to word's starting key
+        final int distanceSquared = wordStartKey.squaredDistanceFrom(corners[0], corners[1]);
```
menny: Micro-optimization:
first check if the `wordStartKey != startKey` and if so then calculate the distance and do the other stuff.

menny: If you are not worried about this, or do not agree, lmk

menny (changes requested):
this is great, I have a few questions

ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
menny: 150 dp depends on the device. Mine will be 450 pixels.
Screenshot 2025-12-15 at 3 37 17 PM

Consider experimenting with different values to see how performance and accuracy changes.

menny:
@doodhout with the merge of https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4554, this PR needs to be rebased with latest `main`, so to fix the conflicts.

Also, do you need help with figuring out the other errors in CI?

menny:
@doodhout I will continue your work in https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4624
Thank you so much for this effort!

doodhout:
> @doodhout I will continue your work in #4624 Thank you so much for this effort!

Hi @menny, sorry for the radio silence, last month has been very intense for me.

If you require assistance, I should be able to provide this during the coming weekend.

Thanks for the feedback and the patience!

menny:
You've done an amazing job with this PR. I'll include some optimizations and such in https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4624
I'll need code review @doodhout if you can

menny:
@doodhout we can close this, right?

PR #4542, opened 2025-11-29
Skip overlay normalization when overlay does not contain colors (instead of assuming it's all black and still normalize)

doodhout:
Only normalize overlay when overlay contains actual colors. When the overlay argument does not contain any colors, don't assume the colors are black and need to be normalized, but instead skip normalization of the overlay and just pass it along as-is.

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/CandidateView.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/keyboards/views/CandidateView.java
@@ -123,9 +123,9 @@ public class CandidateView extends View implements ThemeableChild {
   @VisibleForTesting
   static OverlayData getNormalizedOverlayData(OverlayData overlay) {
     if (overlay.getPrimaryDarkColor() != Color.TRANSPARENT
-            || overlay.getSecondaryTextColor() != Color.TRANSPARENT) {
-      return OverlayDataNormalizer.normalize(overlay, 96, overlay.getPrimaryDarkColor(),
-              overlay.getSecondaryTextColor());
+        || overlay.getSecondaryTextColor() != Color.TRANSPARENT) {
+      return OverlayDataNormalizer.normalize(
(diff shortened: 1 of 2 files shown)


menny:
Thank you for the PR.
Before I review, can you provide an actual case where this happens? Which app etc

Also, can you add a unit test for this change?

menny:
> Before I review, can you provide an actual case where this happens? Which app etc

Oh, I see the bug report. Thanks.

doodhout:
> Also, can you add a unit test for this change?

Thanks for your reply!

I will add unit tests ASAP.

doodhout:
@menny I have a question: to normalize the overlay, you pass PrimaryDarkColor and SecondaryTextColor in CandidateView.setThemeOverlay(), but the normalized overlay contains a PrimaryTextColor and a SecondaryTextColor.

Is it possible that instead of the PrimaryDarkColor, the PrimaryTextColor should get passed to the OverlayDataNormalizer.normalize() method?

menny:
@doodhout  Can you post screenshots?
Comparing current and your suggest

doodhout:
@menny Sorry, in retrospect my question made no sense, so you can ignore it.

The way I think I understand how it works: your overlay normalization feature looks at the contrast between text and background and tries to replace the original text colors with predefined ones when the relative contrast is too low (i.e. when the relative luminance difference is too low).

If my understanding of this is wrong, feel free to correct me.

For completeness' sake, I will try posting screenshots of the before and after change later today.

doodhout:
Screenshot comparison.
Bear in mind that I am using the Lean light theme here.

Before:
image

After:
image

doodhout:
I've made sure to add good test coverage for this change.

Can you re-review this PR?
Thanks!

menny:
> The way I think I understand how it works: your overlay normalization feature looks at the contrast between text and background and tries to replace the original text colors with predefined ones when the relative contrast is too low (i.e. when the relative luminance difference is too low).

This is correct

PR #4624, opened 2026-01-15
Completion of PR #4541

menny:
Completion of https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4541

This PR also include a few optimizations to offset the increased computational effort introduced with https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4541. Read a summary [here](https://github.com/AnySoftKeyboard/AnySoftKeyboard/pull/4624#issuecomment-3758193074)

### DIFF
--- a/ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
+++ b/ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
@@ -57,6 +57,13 @@ public class GestureTypingDetector {
    */
   private static final double END_PROXIMITY_PENALTY_FACTOR = 0.000333;
 
+  /**
+   * Multiplier for the threshold when checking if the gesture ends near the word's last character.
+   * We use a more lenient threshold (5x linear distance = 25x squared distance) for the end key
+   * because users tend to be less precise at the end of a gesture.
+   */
+  private static final int END_DIST_THRESHOLD_MULTIPLIER = 25;
+
   // How far away do two points of the gesture have to be (distance squared)?
   private final int mMinPointDistanceSquared;
 
@@ -402,8 +409,7 @@ public class GestureTypingDetector {
 
   @VisibleForTesting
   static boolean hasEnoughCurvature(final int[] xs, final int[] ys, final int middlePointIndex) {
-    // Calculate the radianValue formed between middlePointIndex, and one point in either
-    // direction
+    // Calculate the radianValue formed between middlePointIndex, and one point in either direction
     final int startPointIndex = middlePointIndex - CURVATURE_NEIGHBORHOOD;
     final int startX = xs[startPointIndex];
     final int startY = ys[startPointIndex];
@@ -478,7 +484,7 @@ public class GestureTypingDetector {
           continue;
         }
 
-        // Quadratic penalty based on squared distance
+        // Quadratic penalty: varies with square of distance (distanceSquared)
         proximityPenalty = distanceSquared * PROXIMITY_PENALTY_FACTOR;
       }
 
@@ -503,7 +509,8 @@ public class GestureTypingDetector {
           final Keyboard.Key endKey = mKeysByCharacter.get(lastChar);
           if (endKey != null) {
             final int endDistanceSquared = endKey.squaredDistanceFrom(gestureEndX, gestureEndY);
-            if (endDistanceSquared > mStartKeyProximityThresholdSquared * 25) {
+            if (endDistanceSquared
+                > mStartKeyProximityThresholdSquared * END_DIST_THRESHOLD_MULTIPLIER) {
               continue; // Gesture ends too far from word's last character
             }
             // Soft penalty for end-key distance (affects ranking, not filtering)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,7 +11,7 @@ alwaysApply: true
 - Prefer not adding new 3rd party dependencies.
 - Do not add 3rd-party dependencies without approval. If you have a solution that requires a new 3rd party, ask for an approval before implementing this solution.
 - Do not use the build task `build` in gradle! This task is usually not compatible with Android builds. Prefer using `assembleDebug`.
-- If you decide to write a temporaty file (for example, a script or storing data) you should put it under the folder `scratch/` which is gitignored.
+- If you decide to write a temporary file (for example, a script or storing data) you should put it under the folder `scratch/` which is gitignored.
 
 # Lint and Format
 


menny:
## Test Failure Analysis

The CI is failing on `tests (app)` with failures in `AnyPopupKeyboardTest`, specifically these tests:
- `testKeyboardPopupSupportsMirrorMultipleRowsNotFullBalanced`
- `testKeyboardPopupSupportsMirrorOneRow` 
- `testKeyboardPopupSupportsMirrorMultipleRowsNotFullNotBalanced`
- `testKeyboardPopupSupportsMirrorMultipleFullRows`
- `testKeyboardPopupSupportsMirrorOneRowNotFull`

**Root cause**: These failures are caused by commit `4d74463b6` from the original PR #4541, which modified `Keyboard.Key.isInside()` to make edges inclusive:

```java
// Changed from < to <= for right and bottom edges
&& (x <= this.x + this.width || (rightEdge && x >= this.x))
&& (y <= this.y + this.height || (bottomEdge && y >= this.y))
```

This change was made to support gesture typing where touches exactly on key edges should be considered "inside" the key. However, it has unintended side effects on popup keyboard layout calculations, causing keys to be positioned incorrectly (X mismatch: expected 0 but was 5).

**Note**: The gesture typing improvements I added in the latest commit work correctly - all gesture typing tests pass. This is a pre-existing issue from the original PR that needs to be addressed.

**Recommendation**: The popup keyboard tests need to be updated to account for the new edge-inclusive behavior, or the `isInside()` logic needs to be refined to handle both use cases correctly.

menny:
# Gesture Proximity Performance Optimizations

This document outlines potential performance optimizations for the gesture typing proximity feature introduced in the `review-pr-4541-improvements` branch.

## Background

The proximity feature allows gesture typing to consider nearby keys when starting a gesture, improving accuracy for imprecise starts. However, this introduces a performance penalty because:

1. Every word requires a `wordStartKey` lookup
2. Words not matching the exact start key require `squaredDistanceFrom` calculation
3. The expensive `calculateDistanceBetweenUserPathAndWord` is called for MORE words than before

---

## 1. Bounding Box Pre-filter

### Description

Before calculating the actual squared distance, use a faster Manhattan distance / bounding box check to quickly reject keys that are obviously too far.

### Implementation Sketch

```java
// Fast rejection using Manhattan distance (no multiplication needed)
int dx = Math.abs(Keyboard.Key.getCenterX(wordStartKey) - corners[0]);
int dy = Math.abs(Keyboard.Key.getCenterY(wordStartKey) - corners[1]);

// If either dimension exceeds threshold, the squared distance definitely will too
// (since threshold^2 > dx^2 + dy^2 requires both dx < threshold and dy < threshold)
if (dx > mStartKeyProximityThreshold || dy > mStartKeyProximityThreshold) {
    continue; // Fast rejection
}

// Only now calculate actual squared distance for penalty
final int distanceSquared = wordStartKey.squaredDistanceFrom(corners[0], corners[1]);
if (distanceSquared > mStartKeyProximityThresholdSquared) {
    continue;
}
```

### Trade-offs

- **Pro**: Cheap integer comparisons vs. multiplication
- **Pro**: Can reject many candidates without full distance calculation
- **Con**: Requires storing non-squared threshold (minor)
- **Con**: Slight code complexity increase
- **Con (STRONG - reason for rejection)**: The `squaredDistanceFrom()` calculation is just 2 multiplications and 1 addition - very cheap on modern CPUs. The pre-filter would add `Math.abs()` calls and comparisons, so the actual performance gain is negligible. Given our other optimizations (start-key grouping reduces iteration to nearby keys only, fail-fast aborts expensive path calculations early), this micro-optimization adds complexity for minimal benefit.

### Estimated Impact

Medium - faster rejection for obviously-far keys

### Status: **NOT IMPLEMENTED** - Rejected due to negligible performance gain

---

## 2. Two-Pass Approach with Early Exit

### Description

Split candidate search into two passes:

1. **Pass 1**: Only exact start key matches (original algorithm behavior)
2. **Pass 2**: Proximity candidates (only if Pass 1 results are insufficient)

If Pass 1 yields enough high-quality candidates, skip Pass 2 entirely.

### Implementation Sketch

```java
public ArrayList<String> getCandidates() {
    mCandidates.clear();
    mCandidateWeights.clear();

    // Pass 1: Exact matches only
    findCandidatesForKey(startKey, /* proximityPenalty */ 0);

    // Early exit if we have enough good candidates
    if (mCandidates.size() >= mMaxSuggestions && hasGoodTopCandidate()) {
        return mCandidates;
    }

    // Pass 2: Check proximity keys (excluding already-checked startKey)
    for (Keyboard.Key nearbyKey : getKeysWithinProximity(corners[0], corners[1])) {
        if (nearbyKey != startKey) {
            double penalty = calculateProximityPenalty(nearbyKey, corners[0], corners[1]);
            findCandidatesForKey(nearbyKey, penalty);
        }
    }

    return mCandidates;
}

private boolean hasGoodTopCandidate() {
    // Heuristic: if top candidate has low weight, we have a good match
    return !mCandidateWeights.isEmpty()
        && mCandidateWeights.get(0) < GOOD_MATCH_THRESHOLD;
}
```

### Trade-offs

- **Pro**: Most gestures start on the correct key, so Pass 1 often suffices
- **Pro**: Preserves full proximity search for edge cases
- **Con**: Need to define "good enough" heuristic
- **Con**: Code restructuring required

### Estimated Impact

High - skips proximity search entirely for most well-executed gestures

---

## 3. Quality Gate After Distance Calculation

### Description

After calculating `distanceFromCurve`, apply a stricter filter for proximity candidates. If the path distance is poor, don't bother considering words from adjacent keys.

### Implementation Sketch

```java
final double distanceFromCurve = calculateDistanceBetweenUserPathAndWord(
    corners, mWordsCorners.get(i + dictionaryWordsCornersOffset));

// Quality gate: proximity candidates need better path match
if (wordStartKey != startKey) {
    // Stricter threshold for non-exact-start words
    if (distanceFromCurve > MINIMUM_DISTANCE_FILTER * 0.3) {
        continue; // Poor path match + wrong start key = skip
    }
}

// Original filter for exact matches
if (distanceFromCurve > MINIMUM_DISTANCE_FILTER) {
    continue;
}
```

### Trade-offs

- **Pro**: Reduces candidates added to the sorted list
- **Pro**: Avoids wasting time on poor proximity matches
- **Con**: May filter out valid corrections in rare cases
- **Con**: Need to tune the quality threshold
- **Con (STRONG - reason for rejection)**: We already penalize proximity candidates via `proximityPenalty` (quadratic penalty based on squared distance from gesture start). Adding a hard filter on top of the existing soft penalty is redundant and harmful to quality. The penalty system allows candidates to compete fairly on their merits - a proximity candidate with excellent path match can still rank highly despite the penalty. A hard gate would exclude potentially valid candidates that the penalty system handles gracefully. This optimization is about gating, not quality improvement, and risks degrading the user experience by removing valid suggestions.

### Estimated Impact

Medium - reduces post-calculation overhead for proximity candidates

### Status: **NOT IMPLEMENTED** - Rejected due to quality concerns

---

## 4. Use Integer Math for Filtering

### Description

Use integer arithmetic for threshold comparisons and filtering. Only convert to double for final weight calculation when the word is a serious candidate.

### Implementation Sketch

```java
// Store threshold as int (already done: mStartKeyProximityThresholdSquared)

// Use int for distance comparison
final int distanceSquared = wordStartKey.squaredDistanceFrom(corners[0], corners[1]);
if (distanceSquared > mStartKeyProximityThresholdSquared) {
    continue; // Integer comparison
}

// Only calculate double penalty for candidates that pass filtering
// Could even defer this until after distanceFromCurve check
final double proximityPenalty = distanceSquared * PROXIMITY_PENALTY_FACTOR;
```

### Additional Optimization

Consider using fixed-point arithmetic for the penalty calculation if profiling shows floating-point as a bottleneck:

```java
// Fixed-point with 1000x scale
final int proximityPenaltyScaled = distanceSquared * 667 / 1000000; // ~0.000667
```

### Trade-offs

- **Pro**: Integer operations are faster than floating-point
- **Pro**: Minimal code change
- **Con**: Limited impact if floating-point isn't the bottleneck

### Estimated Impact

Low-Medium - depends on actual CPU time spent in floating-point ops

---

## 5. Path Length Filter (Limiting Scope)

### Description

Compare the total length of the user's drawn gesture vs. the theoretical path length of the word. If they differ significantly (e.g., > 50%), skip the expensive calculation.

### Implementation Sketch

```java
// Pre-calculate word path lengths (loading phase)
float wordPathLen = mWordPathLengths[i];
// Calculate user path length (once per gesture update)
float userPathLen = mWorkspaceData.getCurrentGestureLength();

if (Math.abs(userPathLen - wordPathLen) / wordPathLen > 0.5f) {
    continue;
}
```

### Trade-offs

- **Pro**: Filters out "short" gestures for "long" words and vice-versa
- **Pro**: Length calculation is cheap (often already available or easy to track)
- **Con**: Users vary wildly in drawing speed/tightness (loops vs straight lines)
- **Con**: "Curvature" logic might make length comparison noisy

### Estimated Impact

Medium - helpful for disambiguating short vs long words

---

## Implementation Priority

| Priority | Optimization               | Effort | Impact     |
| -------- | -------------------------- | ------ | ---------- |
| 1        | Two-Pass with Early Exit   | Medium | High       |
| 2        | Path Length Filter         | Medium | Medium     |
| 3        | Integer Math for Filtering | Low    | Low-Medium |

**Rejected:**
- Bounding Box Pre-filter - see section 1 for rationale
- Quality Gate After Distance - see section 3 for rationale

---

## Applied Optimizations

### Reduce Threshold to 60dp (19c3a91a5)

Reduced proximity threshold from 150dp to 60dp in `dimens.xml`, decreasing the search radius by ~6x.

### Pre-group Words by Starting Key + Memory-efficient Storage (0b4b8e5f5)

Combined optimization that:
1. Groups words by starting key during loading - reduces iteration from O(all words) to O(words for nearby keys)
2. Uses Structure of Arrays (CompactWordList) - reduces memory from ~32 bytes/word to ~12 bytes/word
3. Thread-safe dictionary indexing with `Observable.zipWith(range)`

### End-Key Pruning + CompactWordList Memory Optimization (477e022be)

Combined optimization that:
1. **End-key pruning**: Filters word candidates whose last character's key is too far from gesture end. Uses 5x linear distance threshold (25x squared) - more lenient than start-key because users often overshoot/undershoot at gesture end.
2. **CompactWordList INITIAL_CAPACITY**: Increased from 64 to 256, reducing reallocations from 5 to 3 during loading (~1600 words/key average).
3. **CompactWordList trim()**: Added method to shrink arrays to exact size after loading, reclaiming unused memory from doubling strategy.

### Concurrency Bug Fix: flatMap → concatMap (c70501a2f)

Fixed a critical concurrency bug in `generateCornersInBackground`:
- Changed `flatMap` to `concatMap` to ensure sequential dictionary processing
- `Schedulers.io()` uses an unbounded thread pool, so `flatMap` could process dictionaries in parallel
- This would corrupt shared mutable state (`workspaceData`, `wordsByStartKey`, `wordsCorners`)

### Code Refactoring: Simplify Observable Chain + Extract Helper (2e47d1e2b)

Code cleanup from code review:
1. **Removed CornersGenerationData class**: The `zipWith` pattern was overly complex. Now uses `Observable.range()` directly with lambda capturing method parameters.
2. **Extracted `calculateDirectionPenaltyMultiplier()` helper**: Consolidated duplicated direction penalty calculation logic from both loops in `calculateDistanceBetweenUserPathAndWord()`.

Net reduction of ~40 lines while improving readability.

### Fail-Fast Path Calculation (f735b6adf)

Implemented early termination in `calculateDistanceBetweenUserPathAndWord`:
1. **Added `maxDistance` parameter**: Allows specifying a threshold for early termination
2. **Fail-fast checks in both loops**: Returns `Double.MAX_VALUE` immediately when cumulative distance exceeds threshold
3. **Dynamic threshold in `getCandidates()`**: Uses `Math.min(MINIMUM_DISTANCE_FILTER, worstCandidateWeight)` when we have enough candidates

This prevents wasting CPU cycles computing full distances for words that are already worse than the current worst candidate. As better candidates are found, the threshold tightens progressively, enabling increasingly aggressive pruning.

## Notes

- These optimizations are complementary and can be combined
- Profiling before/after each change is recommended
- Algorithm improvements and background/concurrent work are out of scope for this document

menny:
@doodhout, please review.

ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -326,61 +447,108 @@ public ArrayList<String> getCandidates() {
...
+        // Calculate squared distance from gesture start to this key
+        final int distanceSquared = wordStartKey.squaredDistanceFrom(gestureStartX, gestureStartY);
+
+        // Skip keys that are too far from gesture start
+        if (distanceSquared > mStartKeyProximityThresholdSquared) {
           continue;
         }
 
+        // Quadratic penalty based on squared distance
+        proximityPenalty = distanceSquared * PROXIMITY_PENALTY_FACTOR;
```
doodhout: This doesn't look like it's using the quadratic penalty like it says in the comment above this line.

ime/app/src/main/java/com/anysoftkeyboard/gesturetyping/GestureTypingDetector.java
```diff
@@ -326,61 +447,108 @@ public ArrayList<String> getCandidates() {
...
+        if (word.length > 0) {
+          final char lastChar = Dictionary.toLowerCase(word[word.length - 1]);
+          final Keyboard.Key endKey = mKeysByCharacter.get(lastChar);
+          if (endKey != null) {
+            final int endDistanceSquared = endKey.squaredDistanceFrom(gestureEndX, gestureEndY);
+            if (endDistanceSquared > mStartKeyProximityThresholdSquared * 25) {
+              continue; // Gesture ends too far from word's last character
+            }
+            // Soft penalty for end-key distance (affects ranking, not filtering)
+            endProximityPenalty = endDistanceSquared * END_PROXIMITY_PENALTY_FACTOR;
```
doodhout: Same here.

But then again, I'm wondering about the impact of my suggestion.
Did you already (kind of) finetune these values for optimal results?

doodhout (approved):
You've put a lot of work into improving this further, excellent job!

menny:
Thanks @doodhout . Can you pull this locally and try it out?

doodhout:
@menny Done. I will try this build for a couple of days and report back.

menny:
@doodhout how it feels?

doodhout:
@menny it feels good actually! I would say the current value is good.

PR #4682, opened 2026-02-23
Icelandic keyboard: Improved consistency

Ossur:
Long pressing "a" key now defaults to the letter 'á' and offers 'æ' as the 2nd option instead of defauling to 'à' which doesn't even exist in Icelandic.

Long pressing "t" key now offers the 'þ' letter, which is consistent with that long pressing "d" key offers 'ð'


addons/languages/icelandic/pack/src/main/res/xml/is_qwerty.xml
```diff
@@ -9,7 +9,7 @@
         <Key android:codes="119" android:keyLabel="w" android:popupCharacters="∑"/>
         <Key android:codes="101" android:keyLabel="e" android:popupKeyboard="@xml/is_popup_e" ask:hintLabel="é"/>
         <Key android:codes="114" android:keyLabel="r" android:popupCharacters="®"/>
-        <Key android:codes="116" android:keyLabel="t" android:popupCharacters=""/>
+	<Key android:codes="116" android:keyLabel="t" android:popupCharacters="þ"/>
```
menny: Fix the indentation here

Issue #4727, opened 2026-04-01
Suggest "I'll" when typing "ill" (and more)


highfalutin:
In the English keyboard, I find I have to manually enter apostrophes in many common contractions. Suggestions and auto-corrections would save time in many of these situations:

| Suggest | when typing |
|---------|-------------|
| I'll    | ill         |
| I'd     | id          |
| she'll  | shell       |
| she'd   | shed        |
| he'll   | hell        |
| we'd    | wed         |
| we're   | were        |

_I'd_ be surprised if there aren't others, too... I'm sure _I'll_ come across them later.

@zevrix

faintoldrhyme:
Yes please!! Especially since most available keyboards don't have the apostrophe easily accessible. This would be a game changer.

Issue #4704, opened 2026-03-15
Suggest "it's" when typing "its"


zevrix:
Please suggest "it's" when typing "its". That's an essential suggestion that's expected at all times.

PR #4729, opened 2026-04-01
add autotext for common english contractions

highfalutin:
Add suggestions for common contractions (e.g. "shed" => "she'd"). See issue #4727.


highfalutin:
Added another commit to address #4704.

PR #4734, opened 2026-04-04
addons: iw-rIL strings

elig0n:


menny (changes requested):
Thank you.
We use https://crowdin.com/project/anysoftkeyboard to manage translations. Please, use crowdin to provide translations

PR #4749, opened 2026-05-01
fix: Shift+Enter sends KeyEvent with META_SHIFT_ON instead of commitText("\n")

Honjan:
## Problem
When Shift is held and Enter is pressed, the keyboard sends a plain
newline character via `commitText("\n")`. This loses the shift modifier
information, so apps cannot distinguish Shift+Enter from plain Enter.

## Solution
Send a proper KeyEvent with META_SHIFT_ON for both ACTION_DOWN and
ACTION_UP, consistent with how physical keyboards behave.

## Testing
Tested in SSH/terminal apps where Shift+Enter has a different meaning
than Enter alone.


menny:
You'll need to also fix the unit-test `AnySoftKeyboardGimmicksTest > testSendsENTERCharacterIfShiftIsPressedAndImeHasAction` which ensured the previous behavior is expected.

